PR #6, opened 2023-02-13
Partially fix --device-id

CCRcmcpe:
image

Fixes the error shown above when `--device-id` is set to non-zero.

Note that the current implementation of lowvram is suboptimal to say atleast, it should follow the `--lowvram` CLI flag and be compatible with `--device-id`, and follow `modules.devices` API, I will not fix it in this PR.


CCRcmcpe:
There are still a lot of usage of `.cuda()` especially in annotator part, its use should be discouraged.

PR #29, opened 2023-02-14
Add support for using the web UI torch device

brkirch:
Instead of assuming the default CUDA device is being used, this change will use the torch device from web UI. This is needed if the CUDA device ID web UI is using is not 0, or if the user is running from CPU or MPS (the torch device Macs use).

### DIFF
--- a/annotator/hed/__init__.py
+++ b/annotator/hed/__init__.py
@@ -4,7 +4,7 @@ import torch
 from einops import rearrange
 
 import os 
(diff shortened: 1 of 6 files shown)


brkirch:
I've decided I'll leave the torch device changes for uniformer for a later PR, because a considerable amount of code needs changing.

Mikubill:
Looks good!

PR #36, opened 2023-02-14
When the user does not give an additional image, the init_image is used as the default

fishslot:


Mikubill:
Seems like it will break `StableDiffusionProcessingTxt2Img`, maybe we should add some check for img2img mode.

Issue #46, opened 2023-02-14
Error when using Img2Img Batch


SSA4:
Hello,

I get this error when I try to process multiple images with the img2img batch function:

```
Traceback (most recent call last):
  File "C:\StableDiffusion2023\stable-diffusion-webui-master(3)\stable-diffusion-webui-master\modules\call_queue.py", line 56, in f
   res = list(func(*args, **kwargs))
  File "C:\StableDiffusion2023\stable-diffusion-webui-master(3)\stable-diffusion-webui-master\modules\call_queue.py", line 37, in f
    res = func(*args, **kwargs)
  File "C:\StableDiffusion2023\stable-diffusion-webui-master(3)\stable-diffusion-webui-master\modules\img2img.py", line 163, in img2img
    process_batch(p, img2img_batch_input_dir, img2img_batch_output_dir, img2img_batch_inpaint_mask_dir, args)
  File "C:\StableDiffusion2023\stable-diffusion-webui-master(3)\stable-diffusion-webui-master\modules\img2img.py", line 76, in process_batch
    processed_image.save(os.path.join(output_dir, filename))
AttributeError: 'numpy.ndarray' object has no attribute 'save'
```

I tried to use different preprocessors and models but everytime I get the same error. When I just use img2img with just a single image there is no error. I using Windows 10 whith a NVIDIA RTX 3090 Ti.

Echolink50:
Same

enn-nafnlaus:
I'm not even sure I follow how to use batch with this.  Batch lets you specify a whole directory of images as input images, but this script only lets you specify one image in its canvas.  Shouldn't the script also add the ability to provide a corresponding directory of images for the script to use?  That's what I would need, at least...

SSA4:
When I use this extension I dont put any images in the canvas, so I leave it completly blank. I just enable the 
the ControlNet in the img2img tab and select the preprocessor and the model and set the weight. The 
results are very good and extremely accurate. I get a way better consistency and the results are much better than I get with the normal img2img.

enn-nafnlaus:
Well, that's great if your goal isn't to provide a custom depthmap / etc as distinct from the image...

AbyszOne:
> Well, that's great if your goal isn't to provide a custom depthmap / etc as distinct from the image...

I get the same image without canvas. Thats for scribles i guess.

SSA4:
[...]

PR #56, opened 2023-02-15
fix img2img batch generation

ddPn08:
The solution for #46 .

Add Batch generation checkbox to prevent detection_map from being added to `result`.
This is the process because I couldn't find the variable that determines whether it is batch processing from the pipeline.

### DIFF
--- a/scripts/controlnet.py
+++ b/scripts/controlnet.py
@@ -143,7 +143,8 @@ class Script(scripts.Script):
                     enabled = gr.Checkbox(label='Enable', value=False)
                     scribble_mode = gr.Checkbox(label='Scribble Mode (Reverse color)', value=False)
                     lowvram = gr.Checkbox(label='Low VRAM (8GB or below)', value=False)
-                    
+                    batch_generation = gr.Checkbox(label='Batch generation', value=False, visible=is_img2img)
+
                 ctrls += (enabled,)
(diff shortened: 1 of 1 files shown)


Mikubill:
Added in settings panel: https://github.com/Mikubill/sd-webui-controlnet/commit/4b26a6d6591af3ebc400f70bd1c2cd9451c91646

vjsaw:
If you dont put in an output path, just leave it blank, and let it write in the image2image folder. That works for me.  Before I tried to edit the controlnet.py manually, but after CN doesnt showed up any longer in ui. 😂  Im back to  the original controlnet.py file and leave the output blank. Solution for now 🙂

ljleb:
IIUC, this is related to https://github.com/Mikubill/sd-webui-controlnet/discussions/111#discussioncomment-5000214
> In settings tab, tick the "Do not append map to output" checkbox.

Verte71:
thanks guys "Do not append map to output" works for batch, but it stopped rendering these maps/additional images that are really helpful, and even when i switched it back by unchecking "Do not append map to output" it doesn't work. Any tip to get them back to appear?

ljleb:
@Verte71 as of latest version the code will automatically remove the detectmap when in batch mode. you can now update the extension and safely uncheck the option in the settings.

PR #57, opened 2023-02-15
Allow other scripts to control this extention

fishslot:


Mikubill:
Nice idea. Maybe we should add some hints to tell user params have changed or add a switch to settings panel

fishslot:
You are right,  i add a check box to let the user decide whether to allow other scripts to control the controlnet

Mikubill:
Looks good. Note that I will move the `Enable API` option to settings panel later.

PR #60, opened 2023-02-15
Add `--controlnet-dir PTH_DIR` cli argument

ljleb:
Adds the `--controlnet-dir` cli argument to webui to be able to customize the path to the models. One use case for this is loading control net models from an SSD. I tried keeping the --help text as close to webui's existing options for consistency.

### DIFF
--- a/preload.py
+++ b/preload.py
@@ -1,0 +1,7 @@
+import os.path
+
+EXTENSION_DIR = os.path.dirname(os.path.abspath(__file__))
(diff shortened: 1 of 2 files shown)


ljleb:
Hm, should we keep the `cldm_v15.yaml` file with the models or instead hardcode its location in `extensions/sd-webui-controlnet/models` like it is before this change?

Mikubill:
Good idea but It might be better to move it into something like settings tab.

Mikubill:
Looks good! Will add more options to that panel later.

PR #83, opened 2023-02-16
Some UI adjustments

missionfloyd:
- Moved image to top
- Changed refresh button to match existing
- Relabeled "Open drawing canvas" button to be more accurate

Untitled


Mikubill:
Looks good!

PR #93, opened 2023-02-16
add explanation of batch processing

ddPn08:
Readme changes.
I added this because I have seen many people confused about the Batch process.


ddPn08:
I may have misunderstood a little bit, so I'll cancel it once.

Issue #74, opened 2023-02-15
[Feature Request] Batch process images using the ControlNet image input.


FizzleDorf:
Hello. I was wondering if it was possible to run a batch process using the ControlNet input image rather than using batch img2img so the functionality could also be used in txt2img. I think there could be interesting results especially with animations. Bonus points if you get the ffmpeg commands in there as well so the user doesn't have to stitch everything together. From what I am seeing it can only read a single image for generating both txt2img and img2img.

FizzleDorf:
I've looked through Issue #61. This could be a way to kill two birds with one stone.

FizzleDorf:
I tried to use [Video Loopback](https://github.com/fishslot/video_loopback_for_webui) with ControlNet and since Video Loopback can't do batch img2img, only the first frame is affected. Hardly an implementation if you ask me. In the Loopback discussions I've asked to see if it was possible since there needs to be an input image in order for it to work. Would this still be considered if there isn't any way to get batch img2img for video loopback?

marcsyp:
I was about to open an issue for this, and here are my comments.  I believe this would be a HUGE feature.

---

I have been experimenting with using img2img with a source image and the same image to control all of my controlnet networks vs doing the same exact setup in txt2img (i.e., without the source image as a starting point) -- and I have found the txt2img to provide consistently better results, even after playing with weights, etc, in img2img.

However, txt2img doesn't currently allow you to process images in batch (makes sense). Current workflow to process a large number of images is painful, involving replacing the source image for each control net one by one, clicking generate, and then moving to the next (at least this is less painful using the new tab interface, but still painful).

Proposed workflow
Load a "ControlNet Batch" script
Provide a directory of source images (ideally one directory for each controlnet, up to 10)
For each image, the script replaces the controlnet source image (for each controlnet) with the corresponding image for that directory and runs a txt2img, moves to the next.
Additional Features (Nice to haves)
ability to vary the prompt by index position (directory for text files with different prompts with a way to assign each file to multiple index positions, with the A1111 input as a fallback?)

Another thought:

This may be possible to accomplish in img2img by simply providing a checkbox in the controlnet UI to ignore source image (may need to be a global checkbox?)

Just a few use cases to think about regarding this feature. These particularly related to animation, but could easily be applied to other use cases.

ANIMATION

[...]

Issue #61, opened 2023-02-15
[Feature Request] Support for batch img2img?


Creative-Ataraxia:
how do we load the same image into controlnet that is loaded into img2img? So we dont' have to manually load it twice each time

Mikubill:
Already supported by #36

Creative-Ataraxia:
> Already supported by #36

thanks for the reply; What I meant was, not just the `img2img` tab, but the `batch` subtab of the `img2img` tab; when using the `batch` tab and the controlnet script, will raise error:

```\stable-diffusion-webui\modules\img2img.py", line 76, in process_batch
processed_image.save(os.path.join(output_dir, filename))
AttributeError: 'numpy.ndarray' object has no attribute 'save'```

sounds like the output data of controlnet is a `ndarray` and when processed as webui's code `processed_image`, by line 76, `processed_image.save(os.path.join(output_dir, filename))`, will raise an error

ryanthegray74:
yes currently batch only gives you the very first image and does not work for the rest , this shouldnt be that hard to fix

Creative-Ataraxia:
> yes currently batch only gives you the very first image and does not work for the rest , this shouldnt be that hard to fix

wonder if we can extract the for loop of the batch function and output the image into controlnet somehow

Mikubill:
Added an option to skip saving controlnet detectmep. May help with this issue

boredofnames:
Can confirm this is working now with "Do not append detectmap to output" enabled in settings. Thanks!

toyxyz:
> Added an option to skip saving controlnet detectmep. May help with this issue

When I use "Do not append detectmap to output" Batch img2img works fine! In my personal opinion, this option would be more convenient to use if moved to the ControlNet panel of the img2img tab.

Creative-Ataraxia:
> Added an option to skip saving controlnet detectmep. May help with this issue

great, thanks! just wondering, does this mean that the open pose are no longer guiding the generated images? I noticed that some some output images are not following the input image's pose anymore, with openpose selected

ryanthegray74:
maybe adding more weight can fix that

toyxyz:
[...]

PR #98, opened 2023-02-16
add option not to run img2img when using initial image of img2img tab

ddPn08:
In the img2img tab, added an option not to process img2img when obtaining the input image of ControlNet from the initial image of img2img.
This enables ControlNet batch processing. 

#61 #74


greyphilosophy:
Might just be me, but I'm getting an error testing this:
  File "C:\stable-diffusion-webui\modules\call_queue.py", line 56, in f
    res = list(func(*args, **kwargs))
  File "C:\stable-diffusion-webui\modules\call_queue.py", line 37, in f
    res = func(*args, **kwargs)
  File "C:\stable-diffusion-webui\modules\img2img.py", line 163, in img2img
    process_batch(p, img2img_batch_input_dir, img2img_batch_output_dir, img2img_batch_inpaint_mask_dir, args)
  File "C:\stable-diffusion-webui\modules\img2img.py", line 76, in process_batch
    processed_image.save(os.path.join(output_dir, filename))
AttributeError: 'numpy.ndarray' object has no attribute 'save'

image

FizzleDorf:
I'm not seeing the option. Pulled to latest.

Capture

ddPn08:
@greyphilosophy 
You have to enable "control_net_no_detectmap" in the settings tab.

ddPn08:
@FizzleDorf 
It is displayed only in the img2img tab.  Also, since this is a pull request, the latest version of this repository does not reflect this change.

greyphilosophy:
> @greyphilosophy You have to enable "control_net_no_detectmap" in the settings tab.

Thanks, that was the problem. Is it possible for the script to set it for me and then revert it when it is finished? Or validate that I have the setting correct? Or give me an error message when it fails informing me of the missing setting? Or worst case, a tooltip on the checkbox that tells me it needs that setting enabled? Because right now it's both critical and hard to discover, which isn't a good combination.

ddPn08:
@greyphilosophy Yes, I wish I could, but unfortunately I couldn't figure out how to tell if it's batch processing or not.
I'll think of another way.

greyphilosophy:
@ddPn08 I just retested on a different computer using the main branch and the problem exists there too, so it's not your issue. I've conducted manual testing, and this PR, in my opinion, is working as designed and improves the quality of the extension.

Unfortunately I don't know python so I can't do a code review as well

ljleb:
> @greyphilosophy Yes, I wish I could, but unfortunately I couldn't figure out how to tell if it's batch processing or not. I'll think of another way.

Just letting you know that `HEAD` of main should contain code that allows to detect whether img2img batch tab is selected. You can use [this variable](https://github.com/Mikubill/sd-webui-controlnet/blob/0f18cd87e75c57713b534f3503d5c2a4e3c25362/scripts/controlnet.py#L45) to test whether the batch tab was last active in img2img.

Mikubill:
Looks good but maybe we should consider put this checkbox to batch tab only - for specific use (batch txt2img process).

ddPn08:
@Mikubill 
understood. But I can't think of a way to update the UI on tab change. . . (due to my lack of skill;)
Instead, how about adding items such as "Do not execute img2img during Batch process" on the "Settings" tab?

ljleb:
At the risk of saying something irrelevant again: you could update the css style of the checkbox to disappear in other tabs than img2img batch by setting the `display` property to block or none, depending on whether the tab is active or not. See the `Img2ImgTabTracker` class for img2img tab creation hooks. This seems to be done in some places in `modules/ui.py`.

Another approach could be to just add a checkbox directly inside the img2img batch tab.

Just a couple ideas, wonder if there's a better way to do this.

greyphilosophy:
Doesn't the checkbox work in other img2img modes too though? I'm not sure it needs to be restricted to just batch mode

PR #102, opened 2023-02-16
use relative path for download directory

fur0ut0:
The directory for downloading models are currently hardcoded.
It would be messy when we use a different name for repository directory.

To prevent this, these paths should be expressed as relative path.


Mikubill:
Reverted: incompatible with some os/py version

Issue #121, opened 2023-02-17
Annotator resolution is limited to 1024, and if output set to a higher resolution (for example, 1536), 1024 limit can significantly degrades output quality, especially for "canny" model


Lissanro:
My 8GB video card allows me to generate images up to 1536x1536 and I think higher end card can handle even higher resolution. I suggest to increase Annotator resolution limit to 2048. The same applies to "HED Resolution", "Midas resolution", "Normal resolution", "Hough resolution", which are currently limited to 1024 resolution too.

This is important, because otherwise ControlNet image gets upscaled and this degrades output quality significantly, at least this is the case with canny model and preprocessor. If I want to generate 1536x1536 image, I need to be able to increase Annotator resolution to 1536 to get the expected result. I modified the code and I can confirm that 1536 size works without issues even on 8GB card, this is why I think 2048 could be a good upper limit.

Also, I noticed that if Preprocessor is "none", "Annotator resoultion" slider is disabled by default, which makes sense, but if I choose some other preprocessor, and then "none" again, "Annotator resolution" slider becomes enabled, even though it has no effect on anything as far as I can tell.

Lissanro:
The Pull Request was merged so this issue is resolved.

PR #124, opened 2023-02-17
Increase Annotator resolution to 2048

Lissanro:
Solves issue #121

### DIFF
--- a/scripts/controlnet.py
+++ b/scripts/controlnet.py
@@ -189,56 +189,56 @@ class Script(scripts.Script):
                 def build_sliders(module):
                     if module == "canny":
                         return [
-                            gr.update(label="Annotator resolution", value=512, minimum=64, maximum=1024, step=1, interactive=True),
+                            gr.update(label="Annotator resolution", value=512, minimum=64, maximum=2048, step=1, interactive=True),
                             gr.update(label="Canny low threshold", minimum=1, maximum=255, value=100, step=1, interactive=True),
                             gr.update(label="Canny high threshold", minimum=1, maximum=255, value=200, step=1, interactive=True),
                         ]
                     elif module == "mlsd": #Hough
                         return [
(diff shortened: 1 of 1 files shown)


ljleb:
I see some minimum values set to `minimum=128` instead of `minimum=64`, is that intentional? I don't know about the specifics of the preprocessors, just making sure all values are set correctly.

Lissanro:
Yes, I changed the minimum values to match to the same minimum value in Width and Height sliders in the main section of A1111 UI. I can adjust those in 64-2048 range, and since preprocessor image will be either downscaled or upscaled if it does not match and scaled it can cause issues, this is why it is important to have an option to set Annotator resolution to the same value which is used for output Width/Height. If the Annotator size slider is limited to 1024, there is no way to set the correct value if my output size is for example 1536 or 2048.

Lissanro:
I will update my pull request to resolve the conflict in a moment.

scripts/controlnet.py
```diff
@@ -189,56 +189,56 @@ def refresh_all_models(*inputs):
...
                         return [
-                            gr.update(label="Normal Resolution", value=64, minimum=64, maximum=1024, interactive=False),
+                            gr.update(label="Normal Resolution", value=64, minimum=64, maximum=2048, interactive=False),
                             gr.update(label="Threshold A", value=64, minimum=64, maximum=1024, interactive=False),
                             gr.update(label="Threshold B", value=64, minimum=64, maximum=1024, interactive=False),
                         ]
                     else:
                         return [
-                            gr.update(label="Annotator resolution", value=512, minimum=64, maximum=1024, step=1, interactive=True),
+                            gr.update(label="Annotator resolution", value=64, minimum=64, maximum=2048, interactive=False),
```
Mikubill, line 234: Use `value=512` and `interactive=True` for any other preprocessors might be better

PR #129, opened 2023-02-17
use relative path for download directories ver.2

### LINKED ISSUE #127: The directories for downloading models are hardcoded
For example:

```python
modeldir = os.path.join(extensions.extensions_dir, "sd-webui-controlnet", "annotator", "openpose")
```

It would be messy when we use a different name for repository directory.

To prevent this, these paths should be expressed as relative path.

fur0ut0:
Fix #127.

Fix OS incompatibility issue of #102.

### DIFF
--- a/annotator/hed/__init__.py
+++ b/annotator/hed/__init__.py
@@ -4,7 +4,7 @@ import torch
 from einops import rearrange
 
 import os 
-from modules import extensions, devices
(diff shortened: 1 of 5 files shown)


Mikubill:
already finished in https://github.com/Mikubill/sd-webui-controlnet/pull/138

Issue #45, opened 2023-02-14
[Feature Request] move models and midas to stable-diffusion-webui\models\


brunogcar:
Examples:
https://github.com/thygate/stable-diffusion-webui-depthmap-script  - uses MiDaS and LeReS directly from webui\models\ so  its not needed to download it again, can be shared by multiple extensions
https://github.com/Extraltodeus/depthmap2mask - for midas as well

https://github.com/arenatemp/stable-diffusion-webui-model-toolkit  - dont remember if it downloads something to models, but it extracts components from checkpoints to a folder there
https://github.com/Klace/stable-diffusion-webui-instruct-pix2pix - was already doing that before being integrated into main 

there are some other extensions that either share or create their own folders in models

also ControlNet folder in default path IMO would be better suited

anyway, outstanding job, really well done, thanks

EDIT:
dpt_hybrid-midas-501f0c75 - specificly i have it already in stable-diffusion-webui\models\midas, for depth extensions, aslo:
dpt_beit_large_384.pt
dpt_beit_large_512.pt
dpt_large-midas-2f21e586.pt
midas_v21_small-70d6b9c8.pt
midas_v21-f6b98070.pt

2blackbar:
yeah  weights are now buried in extensions folder when theres a specific models folder for just that so other extensions can use the same weights if needed

abi:
Yeah this would be much better!

ljleb:
Note that there is a webui flag for this at the moment `--controlnet-dir <path-to-pth-dir>`. see #60 (PR also adds a text field in the settings for additional control models directory).

brunogcar:
this extension is probably gonna be the most popular if not already, so the longer it takes, more ppl will have issues when its done and it actually needs to be done.

not just the models in controlsnet, but the ones in anotator subfolders as well, should be in default path models folder,  so other extensions can also use it, some already do it, so waste of space, bandwidth having to download things more than once

actual file structure made logic when it was a standalone project, but now it is in a framework where extensions share common resources , and it actually opened the doors for some other projects, i can see for example eventually someone creating draw by color and using the same model,  etc 

eventually might be integrated to webui, but would be in best practice to use default path and as it is already popular and frankly game changer, set example for future extensions and even inspire future devlopment

PR #138, opened 2023-02-17
Move models to models/ControlNet

missionfloyd:
I wasn't sure what (if anything) to do with the config files, so I left them where they are.

### DIFF
--- a/scripts/controlnet.py
+++ b/scripts/controlnet.py
@@ -15,13 +15,14 @@ from torchvision.transforms import Resize, InterpolationMode, CenterCrop, Compos
 from scripts.cldm import PlugableControlModel
 from scripts.processor import *
 from modules.ui_components import ToolButton
+from modules.paths import models_path
 
 CN_MODEL_EXTS = [".pt", ".pth", ".ckpt", ".safetensors"]
 cn_models = {}      # "My_Lora(abcd1234)" -> C:/path/to/model.safetensors
 cn_models_names = {}  # "my_lora" -> "My_Lora(abcd1234)"
-cn_models_dir = os.path.join(scripts.basedir(), "models")
+cn_models_dir = os.path.join(models_path, "ControlNet")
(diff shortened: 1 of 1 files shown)


ljleb:
I think it would be a good idea to additionally update the readme, as otherwise the guidance on where to place the models will be misleading.

scripts/controlnet.py
```diff
@@ -35,10 +36,10 @@
 CN_MODEL_EXTS = [".pt", ".pth", ".ckpt", ".safetensors"]
 cn_models = {}      # "My_Lora(abcd1234)" -> C:/path/to/model.safetensors
 cn_models_names = {}  # "my_lora" -> "My_Lora(abcd1234)"
-cn_models_dir = os.path.join(scripts.basedir(), "models")
+cn_models_dir = os.path.join(models_path, "ControlNet")
 os.makedirs(cn_models_dir, exist_ok=True)
+default_conf = os.path.join(scripts.basedir(), "models", "cldm_v15.yaml")
```
Mikubill, line 41: For backward compatibility, we should also consider the original model path in case some one need it

missionfloyd: Okay, I added the old path as well.

brunogcar:
missing leres, that was added yesterday

missionfloyd:
I was wondering if there was a better place for them. If separate folders makes more sense, let's do that.

brunogcar:
those models predate ControlNet , that file structure made logic when it was a standalone project, but now its on a framework that shares resources, IMO now makes more sense to place them in root models folder

we all know how powerful controlnet is and how it will influence future development, so it would be a good chance to set standards and examples

newest commit seems good to me, that closes https://github.com/Mikubill/sd-webui-controlnet/issues/45

missionfloyd:
@Mikubill If the paths look good, would you kindly merge this so I can be done with it?

Mikubill:
Also close #127 #129. Thanks for your contribution!

Issue #139, opened 2023-02-17
Support batch workflow for image processing


ebceu4:
Currently, the image processing only allows for processing a single image at a time with ControlNet, trying to use batch tab but getting AttributeError: 'numpy.ndarray' object has no attribute 'save'

`
  File "/workspace/stable-diffusion-webui/modules/call_queue.py", line 56, in f
    res = list(func(*args, **kwargs))
  File "/workspace/stable-diffusion-webui/modules/call_queue.py", line 37, in f
    res = func(*args, **kwargs)
  File "/workspace/stable-diffusion-webui/modules/img2img.py", line 163, in img2img
    process_batch(p, img2img_batch_input_dir, img2img_batch_output_dir, img2img_batch_inpaint_mask_dir, args)
  File "/workspace/stable-diffusion-webui/modules/img2img.py", line 76, in process_batch
    processed_image.save(os.path.join(output_dir, filename))
AttributeError: 'numpy.ndarray' object has no attribute 'save'
`

ljleb:
See https://github.com/Mikubill/sd-webui-controlnet/discussions/111
> In settings tab, tick the "Do not append map to output" checkbox.

PR #142, opened 2023-02-17
automatically remove detectmap in img2img batch tab

ljleb:
Automatically removes the detectmap from the output images when the img2img batch tab is active. Also removes the `control_net_no_detectmap` in ui setting as it is no more needed.

I'm seeing multiple reports of `AttributeError: 'numpy.ndarray' object has no attribute 'save'`, for example: #46, #119, #134, #139. I'm thinking this should be a more user friendly implementation. If for some reason people still want to include the detectmap in batch processing, it will need additional code.

### DIFF
--- a/scripts/controlnet.py
+++ b/scripts/controlnet.py
@@ -436,7 +436,8 @@ class Script(scripts.Script):
         
     def postprocess(self, p, processed, *args):
         is_img2img = issubclass(type(p), StableDiffusionProcessingImg2Img)
-        if self.latest_network is None or (is_img2img and active_img2img_tab == 'img2img_batch_tab'):
+        no_detectmap_opt = shared.opts.data.get("control_net_no_detectmap", False)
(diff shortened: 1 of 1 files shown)


ljleb:
Added back the option in case people don't want to see it in the output.
Let me know if we should remove the option after all and I'll revert the last commit in this PR.

Mikubill:
Nice fix.

gitpulled:
I was hoping to extract the skeleton as an openpose figure from a batch file, am i going up the wrong avenue? 

disabling this feature seems to disable what i was trying to do.

Enferlain:
@gitpulled did you figure out how to do that since? I've been trying all day

ljleb:
You can enable detectmap auto saving instead of displaying them in the gallerie if I'm not mistaken. Check your settings, there should be a checkbox and a path to specify where to save.

Enferlain:
Yeah I do that, it's just the openpose controlnet doesn't want function for me for doing batch detectmap saving

PR #176, opened 2023-02-18
Just suggestion: add option for depth map directly

axilesoft:


Mikubill:
Maybe duplicates with annotator "none".

PR #194, opened 2023-02-18
added an API layer

sangww:
This API layer replicates SD-WebUI's txt2img and img2img pipeline, with added script parameters for ControlNet.

It has two caveats that needs addressed, but currently in a working state with full support of SD-WebUI + ControlNet features.

- I still don't know what's the first arg of p.script_args: I observed it to be 0. If someone can validate this!
- This api layer soon need to be able to automatically pull in other AlwaysVisible scripts' parameters, and only update those for ControlNet. Any advice appreciated!


kft334:
Can anyone confirm that Text-To-Image works? I'm getting bad outputs that seem like normal Text-To-Image. Everything loads and there are no errors though. And could you also return the generated control image at either the first or last image index if the pre-processor was used?

sangww:
"controlnet_input_image": [img],
"controlnet_module": 'openpose',
"controlnet_model": 'control_sd15_openpose [fef5e48e]',

This is how json objects are formated and those that are necessary. My setup works fine this way with depth, pose and scribble. Img is a base64 string for the image that goes into the ControlNet UI on the SD WebUI.

Mikubill:
Good. Small advice:

1. Move API stuff to another file, like `api.py`
2. Maybe we should add some examples in the readme.

kft334:
I see the issue now. Doesn't work with Euler apparently? Switched to DDIM and I'm getting proper results now.

chrisbward:
I've dropped this change in, but seems to break when visiting the docs; http://127.0.0.1:7860/docs#/

sangww:
For those this doesn't work, forgot the highlight it only works only when ControlNet is installed as a script in img2img and txt2img. Try it without other scripts installed in the extension tab.

sangww:
> Good. Small advice:
> 
> 1. Move API stuff to another file, like `api.py`
> 2. Maybe we should add some examples in the readme.

Will do when I have time. Quick Q:

I see in your script, that you arrange script args based on script index. Any tips in populating script args for multiple alwayson_scripts properly? Thanks!

sangww:
> For those this doesn't work, forgot the highlight it only works only when ControlNet is installed as a script in img2img and txt2img. Try it without other scripts installed in the extension tab.

@kft334 this might be the issue

kiriri:
Creating a new API was a wonderful idea, even if it's 'makeshift'. It makes things so much easier to edit.

One thing though, I pass in the controlnet_input_image as an array of  base64 strings, and it throws an error ("Incorrect Padding").

Replacing 
```python
cn_image = Image.open(io.BytesIO(base64.b64decode(controlnet_input_image[0])))
 ```
with
```python
cn_image = decode_base64_to_image(controlnet_input_image[0])   
```
Fixed that error. Hope that helps everyone else who's having that issue.

stassius:
Hello! I'm the author of the post on Reddit. I have a couple of suggestions to make your api layer better.

1. You should add GET methods to retrieve the available Preprocessors and Models, just like in the standard A1111 api. 
2. It's better to make it in a different file. It will be easier to maintain.

I still don't know what would be the best solution from the architectural standpoint. You cloned the whole A1111 api for txt2img and img2img. I don't think it's a good idea, as it's hard to maintain. You'll have to track changes both in A1111 api and in the ControlNet extension. My solution is not great either. I made the api only to retrieve the models, and created an A1111 script to control the ControlNet extension, so a user only have to add the Script name and arguments to the existing API call. It's hacky too (and adds an extra script with UI), but at least, you don't have to worry about maintaining the cloned API.  I don't know if there is another way to do this apart from adding the ControlNet to the core of A1111. Any suggestions would be appreciated.

sangww:
> Hello! I'm the author of the post on Reddit. I have a couple of suggestions to make your api layer better.
> 
> 1. You should add GET methods to retrieve the available Preprocessors and Models, just like in the standard A1111 api.
> 2. It's better to make it in a different file. It will be easier to maintain.
> 
> I still don't know what would be the best solution from the architectural standpoint. You cloned the whole A1111 api for txt2img and img2img. I don't think it's a good idea, as it's hard to maintain. You'll have to track changes both in A1111 api and in the ControlNet extension. My solution is not great either. I made the api only to retrieve the models, and created an A1111 script to control the ControlNet extension, so a user only have to add the Script name and arguments to the existing API call. It's hacky too (and adds an extra script with UI), but at least, you don't have to worry about maintaining the cloned API. I don't know if there is another way to do this apart from adding the ControlNet to the core of A1111. Any suggestions would be appreciated.

Hey, I really loved your work! And great suggestions : )

Yeah definitely this is a super makeshift effort : ) I just needed quick and dirty for the moment, but as you said maintaining will be a headache potentially. I think adding the 1) GET methods, 2) move to separate script (done on my system locally), 3) returning control results as you and others suggested would be good for the moment. But also a fan of exploring the right implementation model (which would be why I wanted to start this thread) that would be needed going forward.

sangww:
> @sangww - having problems running this, the fallback case line 350 always is True. therefore control net is not used? Am I missing something?

Shouldn't be true. I have updated the commit which also includes an example api call from a python notebook. This could be helpful in comparing with your setup There are a few things to check given how makeshift is this:
- Do you have other scripts installed on txt2img and img2img? Should have none for this to correctly work in the current implementation. You could simply disable those (such as additional network)
- Do you see any error message? I wonder if the input image for controlnet is correctly encoded, which could lead to a failed processing.

sangww:
9

> No error message, no extra modules, installed your branch on a fresh install - I use the same encoding for the input for ControlNet as the initImage (without the "data:image/png;base64," substring) and still the flag is checked True.
> 
> Could you double check if this is occurring for you too? -> There is no exception clause in the chain of execution afterward - whence there is no error displayed.

That is strange indeed. I am suspecting it could be the extension's own seeting.
Could you try: go to settings tab -> ControlNet -> ensure true is set for "Allow other script to control this extension"?

Ericxgao:
I have the same issue still, with the settings changes.

Mikubill:
add more entries to script_args should fix it. (PR welcome
https://github.com/Mikubill/sd-webui-controlnet/blob/5fed282f93ef28952dc517b0bb2e8dc3a1cca278/scripts/api.py#L170-L180

medledan:
> I see the issue now. Doesn't work with Euler apparently? Switched to DDIM and I'm getting proper results now.
> 
> Edit: Just tried with DPM++ 2M and the output was bad as well. Using canny btw
> 
> Edit: Just tested with depth and the same thing happens. It works with DDIM and PLMS but with Euler, LMS, Heun or DPM++ 2M the results become random. I've tried most of the samplers now and ControlNet appears to only be applying to DDIM and PLMS. They all work in the webui though. It could just be an issue with my installation.

This is why I wait like 1-2 weeks after a commit to update anything. Thanks for the comment. I'll wait a while longer before updating.

bigahega:
> I have the same issue still, with the settings changes.
> 
> ```
>         # todo: extend to include wither alwaysvisible scripts
>         processed = scripts.scripts_img2img.run(p, *(p.script_args))
> 
>         if processed is None:  # fall back
>             processed = process_images(p)
> ```
> 
> Processed is always none here

I am also facing the same issue. After a quick look, it seems like `processed` is None because `script_index == 0` in script.run(). Launching with --nowebui btw.

chrisbward:
Screenshot from 2023-02-21 05-27-17

This seems incomplete? Can't see where the SD Model is set?

chrisbward:
> 

+1 on GET request for retrieving all preprocessors and controlnet models

YuanBo66668888:
> > I have the same issue still, with the settings changes.
> > ```
> >         # todo: extend to include wither alwaysvisible scripts
> >         processed = scripts.scripts_img2img.run(p, *(p.script_args))
> > 
> >         if processed is None:  # fall back
> >             processed = process_images(p)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Processed is always none here
> 
> I am also facing the same issue. After a quick look, it seems like `processed` is None because `script_index == 0` in script.run(). Launching with --nowebui btw.

Processed is always none here, What should I do

YuanBo66668888:
My problem is as follows:
`ERROR:    Exception in ASGI application
  File "/home/xxxx/xx/stable-diffusion-webui/venv/lib/python3.8/site-packages/anyio/streams/memory.py", line 94, in receive
    return self.receive_nowait()
  File "/home/xxxx/xx/stable-diffusion-webui/venv/lib/python3.8/site-packages/anyio/streams/memory.py", line 89, in receive_nowait
    raise WouldBlock
anyio.WouldBlock

During handling of the above exception, another exception occurred:

  File "/home/xxxx/xx/stable-diffusion-webui/venv/lib/python3.8/site-packages/starlette/middleware/base.py", line 77, in call_next
    message = await recv_stream.receive()
  File "/home/xxxx/xx/stable-diffusion-webui/venv/lib/python3.8/site-packages/anyio/streams/memory.py", line 114, in receive
    raise EndOfStream
anyio.EndOfStream

During handling of the above exception, another exception occurred:

  File "/home/xxxx/xx/stable-diffusion-webui/venv/lib/python3.8/site-packages/uvicorn/protocols/http/h11_impl.py", line 407, in run_asgi
    result = await app(  # type: ignore[func-returns-value]
  File "/home/xxxx/xx/stable-diffusion-webui/venv/lib/python3.8/site-packages/uvicorn/middleware/proxy_headers.py", line 78, in __call__
    return await self.app(scope, receive, send)
  File "/home/xxxx/xx/stable-diffusion-webui/venv/lib/python3.8/site-packages/fastapi/applications.py", line 271, in __call__
    await super().__call__(scope, receive, send)
  File "/home/xxxx/xx/stable-diffusion-webui/venv/lib/python3.8/site-packages/starlette/applications.py", line 125, in __call__
    await self.middleware_stack(scope, receive, send)
  File "/home/xxxx/xx/stable-diffusion-webui/venv/lib/python3.8/site-packages/starlette/middleware/errors.py", line 184, in __call__
    raise exc
  File "/home/xxxx/xx/stable-diffusion-webui/venv/lib/python3.8/site-packages/starlette/middleware/errors.py", line 162, in __call__
    await self.app(scope, receive, _send)
  File "/home/xxxx/xx/stable-diffusion-webui/venv/lib/python3.8/site-packages/starlette/middleware/base.py", line 104, in __call__
    response = await self.dispatch_func(request, call_next)
  File "/home/xxxx/xx/stable-diffusion-webui/modules/api/api.py", line 96, in log_and_time
    res: Response = await call_next(req)
  File "/home/xxxx/xx/stable-diffusion-webui/venv/lib/python3.8/site-packages/starlette/middleware/base.py", line 80, in call_next
    raise app_exc
  File "/home/xxxx/xx/stable-diffusion-webui/venv/lib/python3.8/site-packages/starlette/middleware/base.py", line 69, in coro
    await self.app(scope, receive_or_disconnect, send_no_error)
  File "/home/xxxx/xx/stable-diffusion-webui/venv/lib/python3.8/site-packages/starlette/middleware/gzip.py", line 24, in __call__
    await responder(scope, receive, send)
  File "/home/xxxx/xx/stable-diffusion-webui/venv/lib/python3.8/site-packages/starlette/middleware/gzip.py", line 44, in __call__
    await self.app(scope, receive, self.send_with_gzip)
  File "/home/xxxx/xx/stable-diffusion-webui/venv/lib/python3.8/site-packages/starlette/middleware/exceptions.py", line 79, in __call__
    raise exc
  File "/home/xxxx/xx/stable-diffusion-webui/venv/lib/python3.8/site-packages/starlette/middleware/exceptions.py", line 68, in __call__
    await self.app(scope, receive, sender)
  File "/home/xxxx/xx/stable-diffusion-webui/venv/lib/python3.8/site-packages/fastapi/middleware/asyncexitstack.py", line 21, in __call__
    raise e
  File "/home/xxxx/xx/stable-diffusion-webui/venv/lib/python3.8/site-packages/fastapi/middleware/asyncexitstack.py", line 18, in __call__
    await self.app(scope, receive, send)
  File "/home/xxxx/xx/stable-diffusion-webui/venv/lib/python3.8/site-packages/starlette/routing.py", line 706, in __call__
    await route.handle(scope, receive, send)
  File "/home/xxxx/xx/stable-diffusion-webui/venv/lib/python3.8/site-packages/starlette/routing.py", line 276, in handle
    await self.app(scope, receive, send)
  File "/home/xxxx/xx/stable-diffusion-webui/venv/lib/python3.8/site-packages/starlette/routing.py", line 66, in app
    response = await func(request)
  File "/home/xxxx/xx/stable-diffusion-webui/venv/lib/python3.8/site-packages/fastapi/routing.py", line 237, in app
    raw_response = await run_endpoint_function(
  File "/home/xxxx/xx/stable-diffusion-webui/venv/lib/python3.8/site-packages/fastapi/routing.py", line 163, in run_endpoint_function
    return await dependant.call(**values)
  File "/home/xxxx/xx/stable-diffusion-webui/extensions/sd-webui-controlnet/scripts/api.py", line 220, in txt2img
    processed = process_images(p)
  File "/home/xxxx/xx/stable-diffusion-webui/modules/processing.py", line 486, in process_images
    res = process_images_inner(p)
  File "/home/xxxx/xx/stable-diffusion-webui/modules/processing.py", line 657, in process_images_inner
    x_sample = modules.face_restoration.restore_faces(x_sample)
  File "/home/xxxx/xx/stable-diffusion-webui/modules/face_restoration.py", line 19, in restore_faces
    return face_restorer.restore(np_image)
  File "/home/xxxx/xx/stable-diffusion-webui/modules/codeformer_model.py", line 88, in restore
    self.create_models()
  File "/home/xxxx/xx/stable-diffusion-webui/modules/codeformer_model.py", line 65, in create_models
    checkpoint = torch.load(ckpt_path)['params_ema']
  File "/home/xxxx/xx/stable-diffusion-webui/modules/safe.py", line 106, in load
    return load_with_extra(filename, extra_handler=global_extra_handler, *args, **kwargs)
  File "/home/xxxx/xx/stable-diffusion-webui/modules/safe.py", line 151, in load_with_extra
    return unsafe_torch_load(filename, *args, **kwargs)
  File "/home/xxxx/xx/stable-diffusion-webui/venv/lib/python3.8/site-packages/torch/serialization.py", line 777, in load
    with _open_zipfile_reader(opened_file) as opened_zipfile:
  File "/home/xxxx/xx/stable-diffusion-webui/venv/lib/python3.8/site-packages/torch/serialization.py", line 282, in __init__
    super(_open_zipfile_reader, self).__init__(torch._C.PyTorchFileReader(name_or_buffer))
RuntimeError: PytorchStreamReader failed reading zip archive: failed finding central directory
`
What's wrong?? run with --disable-safe-unpickle --nowebui

chrisbward:
Screenshot from 2023-02-21 05-27-17

No option to set SD model?

chrisbward:
ah okay, I think I have to call /sdapi/v1/options first to set the model, thanks!

sangww:
> ah okay, I think I have to call /sdapi/v1/options first to set the model, thanks!

https://github.com/Mikubill/sd-webui-controlnet/blob/main/example/test_api.ipynb
In fact, you could use sampler index item to set the model in run time. It works the same as sampler_name.

sangww:
> --nowebui

It is hard to see from the log what the issue is. Looks like .pth/.ckpt file issue roughly. Does it run when you use control net via GUI?

entmike:
I was JUST looking for this feature, I appreciate all the fast hard work!  Can't wait to use this API.

sangww:
> As mentioned in my previous message there seems to be missing arguments that unless included cause the execution chain to fail (fallback = True) - Please take a look at this @sangww, it seems that either your setup has a setting enabled - allowing execution - or that the script has some errors. 
> 
> 
> 
> It’d be worth to go down the entire chain of execution to find the cause - I suspect it comes down to some odd continue clause that doesn’t show errors in the log but returns without exceptions

I am seeing strange interactions that I didnt see in my original setup after getting the latest versions of this pull request. It got a bit out of my hand, but one thing I def noticed is this doesn't work with --nowebui cmd arg. (also @YuanBo66668888). What is your command args looking like?

YuanBo66668888:
> > As mentioned in my previous message there seems to be missing arguments that unless included cause the execution chain to fail (fallback = True) - Please take a look at this @sangww, it seems that either your setup has a setting enabled - allowing execution - or that the script has some errors.
> > It’d be worth to go down the entire chain of execution to find the cause - I suspect it comes down to some odd continue clause that doesn’t show errors in the log but returns without exceptions
> 
> I am seeing strange interactions that I didnt see in my original setup after getting the latest versions of this pull request. It got a bit out of my hand, but one thing I def noticed is this doesn't work with --nowebui cmd arg. (also @YuanBo66668888). What is your command args looking like?

Mine is OK. After looking at the code, the main problem is the nowebui situation. It should be that the controlnet process is not running

sangww:
added a new pr #278 that reflects some updates on the main, incorrect args added during the merge, fixed some suspicious parts from my original code, and added network restore behavior for controlnet to not affect subsequent non-controlnet generations.

For those having issue, please let me know. I won't be fast but be monitoring.

bigahega:
Still not working for me with --nowebui. Probably A1111 bug btw.

In api only mode ScriptRunner.initialize_scripts() is only called when a script is executed through /sdapi/v1/txt2img or /sdapi/v1/img2img endpoints by passing a script_name. However, since this extension defines its own endpoints, scripts are never initialized.

sangww:
> Still not working for me with --nowebui. Probably A1111 bug btw.
> 
> 
> 
> In api only mode ScriptRunner.initialize_scripts() is only called when a script is executed through /sdapi/v1/txt2img or /sdapi/v1/img2img endpoints by passing a script_name. However, since this extension defines its own endpoints, scripts are never initialized.

I don't plan on making change for nowebui flag since i don't have much understanding of script lifecycle. It would be great if someone could look into it who knows how scripts are initialized.

kiriri:
I am running into fatal errors when I combine "interrupt" calls with this api.
Eg controlnet/img2img, interrupt, controlnet/img2img, causes the following error :

```
ERROR:    Exception in ASGI application
  File "/home/moot/Desktop/auto1111/stable-diffusion-webui/venv/lib/python3.10/site-packages/anyio/streams/memory.py", line 94, in receive
    return self.receive_nowait()
  File "/home/moot/Desktop/auto1111/stable-diffusion-webui/venv/lib/python3.10/site-packages/anyio/streams/memory.py", line 89, in receive_nowait
    raise WouldBlock
anyio.WouldBlock

During handling of the above exception, another exception occurred:

  File "/home/moot/Desktop/auto1111/stable-diffusion-webui/venv/lib/python3.10/site-packages/starlette/middleware/base.py", line 77, in call_next
    message = await recv_stream.receive()
  File "/home/moot/Desktop/auto1111/stable-diffusion-webui/venv/lib/python3.10/site-packages/anyio/streams/memory.py", line 114, in receive
    raise EndOfStream
anyio.EndOfStream

During handling of the above exception, another exception occurred:

  File "/home/moot/Desktop/auto1111/stable-diffusion-webui/venv/lib/python3.10/site-packages/uvicorn/protocols/http/h11_impl.py", line 407, in run_asgi
    result = await app(  # type: ignore[func-returns-value]
  File "/home/moot/Desktop/auto1111/stable-diffusion-webui/venv/lib/python3.10/site-packages/uvicorn/middleware/proxy_headers.py", line 78, in __call__
    return await self.app(scope, receive, send)
  File "/home/moot/Desktop/auto1111/stable-diffusion-webui/venv/lib/python3.10/site-packages/fastapi/applications.py", line 271, in __call__
    await super().__call__(scope, receive, send)
  File "/home/moot/Desktop/auto1111/stable-diffusion-webui/venv/lib/python3.10/site-packages/starlette/applications.py", line 125, in __call__
    await self.middleware_stack(scope, receive, send)
  File "/home/moot/Desktop/auto1111/stable-diffusion-webui/venv/lib/python3.10/site-packages/starlette/middleware/errors.py", line 184, in __call__
    raise exc
  File "/home/moot/Desktop/auto1111/stable-diffusion-webui/venv/lib/python3.10/site-packages/starlette/middleware/errors.py", line 162, in __call__
    await self.app(scope, receive, _send)
  File "/home/moot/Desktop/auto1111/stable-diffusion-webui/venv/lib/python3.10/site-packages/starlette/middleware/base.py", line 104, in __call__
    response = await self.dispatch_func(request, call_next)
  File "/home/moot/Desktop/auto1111/stable-diffusion-webui/modules/api/api.py", line 96, in log_and_time
    res: Response = await call_next(req)
  File "/home/moot/Desktop/auto1111/stable-diffusion-webui/venv/lib/python3.10/site-packages/starlette/middleware/base.py", line 80, in call_next
    raise app_exc
  File "/home/moot/Desktop/auto1111/stable-diffusion-webui/venv/lib/python3.10/site-packages/starlette/middleware/base.py", line 69, in coro
    await self.app(scope, receive_or_disconnect, send_no_error)
  File "/home/moot/Desktop/auto1111/stable-diffusion-webui/venv/lib/python3.10/site-packages/starlette/middleware/gzip.py", line 24, in __call__
    await responder(scope, receive, send)
  File "/home/moot/Desktop/auto1111/stable-diffusion-webui/venv/lib/python3.10/site-packages/starlette/middleware/gzip.py", line 44, in __call__
    await self.app(scope, receive, self.send_with_gzip)
  File "/home/moot/Desktop/auto1111/stable-diffusion-webui/venv/lib/python3.10/site-packages/starlette/middleware/exceptions.py", line 79, in __call__
    raise exc
  File "/home/moot/Desktop/auto1111/stable-diffusion-webui/venv/lib/python3.10/site-packages/starlette/middleware/exceptions.py", line 68, in __call__
    await self.app(scope, receive, sender)
  File "/home/moot/Desktop/auto1111/stable-diffusion-webui/venv/lib/python3.10/site-packages/fastapi/middleware/asyncexitstack.py", line 21, in __call__
    raise e
  File "/home/moot/Desktop/auto1111/stable-diffusion-webui/venv/lib/python3.10/site-packages/fastapi/middleware/asyncexitstack.py", line 18, in __call__
    await self.app(scope, receive, send)
  File "/home/moot/Desktop/auto1111/stable-diffusion-webui/venv/lib/python3.10/site-packages/starlette/routing.py", line 706, in __call__
    await route.handle(scope, receive, send)
  File "/home/moot/Desktop/auto1111/stable-diffusion-webui/venv/lib/python3.10/site-packages/starlette/routing.py", line 276, in handle
    await self.app(scope, receive, send)
  File "/home/moot/Desktop/auto1111/stable-diffusion-webui/venv/lib/python3.10/site-packages/starlette/routing.py", line 66, in app
    response = await func(request)
  File "/home/moot/Desktop/auto1111/stable-diffusion-webui/venv/lib/python3.10/site-packages/fastapi/routing.py", line 237, in app
    raw_response = await run_endpoint_function(
  File "/home/moot/Desktop/auto1111/stable-diffusion-webui/venv/lib/python3.10/site-packages/fastapi/routing.py", line 163, in run_endpoint_function
    return await dependant.call(**values)
  File "/home/moot/Desktop/auto1111/stable-diffusion-webui/extensions/sd-webui-controlnet/scripts/api.py", line 358, in img2img
    processed = process_images(p)
  File "/home/moot/Desktop/auto1111/stable-diffusion-webui/modules/processing.py", line 486, in process_images
    res = process_images_inner(p)
  File "/home/moot/Desktop/auto1111/stable-diffusion-webui/modules/processing.py", line 713, in process_images_inner
    extra_networks.deactivate(p, extra_network_data)
UnboundLocalError: local variable 'extra_network_data' referenced before assignment

```

This error does not occur with sdapi/v1/img2img. Any ideas how to fix this? Is the api missing some guards?

joseph-schwartz:
Is there API support for: "Multi-ControlNet / Joint Conditioning (Experimental)?"  Been looking for how to do it.

vinch00:
The API seems to be working, except for controlnet_normalmap.
The generated normalmap through the API is a blank purple image.
Which means the generated output image is not using the normalmap.
On the webui, the normalmap loads just fine.
Any ideas?

Mikubill:
If API params passed correctly, they should be handled by main `process()` func: fallback to default val when not set or p_res <=64. Maybe some params' loc is incorrect.
https://github.com/Mikubill/sd-webui-controlnet/blob/f94b453d10c2eb4e3ad57b7cdfdd17e60a557b3e/scripts/controlnet.py#L575-L580

Mikubill:
maybe we should consider using setattr(p, ...) like script control, this way should avoid such problem and robust:

https://github.com/Mikubill/sd-webui-controlnet/blob/f94b453d10c2eb4e3ad57b7cdfdd17e60a557b3e/scripts/controlnet.py#L508-L520

Mikubill:
fixed in https://github.com/Mikubill/sd-webui-controlnet/commit/3691d1130b03823cb3e9fe778a533a5ffb633aac. 
checkout https://github.com/Mikubill/sd-webui-controlnet/blob/main/example/api_img2img.ipynb for examples

Mikubill:
Could you share more details about difference on images/settings?
Note that canny's default val is low(thr_a)=100, high(the_b)=200

Mikubill:
Tested with https://github.com/Mikubill/sd-webui-controlnet/commit/e60e292799f959c46a843001451f339eef269678.
I don't see any difference between different type of input params
| Default | Int(100, 200) | Float(100.0, 200.0) |
|:-------------------------:|:-------------------------:|:-------------------------:|
|   |  |  |

As for aliasing, could you replicate this on normal WebUI? Also share config and params if possible.

shangdibufashi:
is there a way to apply mutil-controlnet?

whitepapercg:
I tried to run Controlnet directly from my bot, but it generates a image using the input data, but not Controlnet. What is wrong with my request?

Mikubill:
Note that you need to enable `Allow other script to control this extension` in settings for API

whitepapercg:
> Note that you need to enable `Allow other script to control this extension` in settings for API access

Thanks you so much! Its work now!

whitepapercg:
> Note that you need to enable `Allow other script to control this extension` in settings for API access

Nowhere did I find information on how to call Controlnet modules #1, #2, #3, etc. through the api, if they are enabled in the settings. How can I specify other modules in "payload"?
I would like to try the "depth" and "canny" combination through my bot

ljleb:
> is there a way to apply multi-controlnet?

Working on it in #384

chrisbward:
Just started getting this error;
```
Error running process_batch: /home/user/stable-diffusion-webui/extensions/sd-webui-additional-networks/scripts/additional_networks.py
  File "/home/user/stable-diffusion-webui/modules/scripts.py", line 395, in process_batch
    script.process_batch(p, *script_args, **kwargs)
  File "/home/user/stable-diffusion-webui/extensions/sd-webui-additional-networks/scripts/additional_networks.py", line 158, in process_batch
    if not args[0]:
IndexError: tuple index out of range
```

I've installed the LoRA network today?

chrisbward:
Hey, so I had to disable "additional networks" extension and it started working again.

Another issue, GET request on "controlnet/model_list" returns an empty array?

marcsyp:
Is there any documentation of how to use the controlnet api other than what is in the A1111 api documentation?  It would be great to see an example of successfully running a txt2img and an img2img using controlnet with POST requests.  I've used POST for regular img2img from a 3d software environement and would like to add the capability to use CN as well.  Thanks!

ljleb:
We definitely need to describe what each field of `controlnet_units` objects expects as input somewhere.

In the meantime, tl;dr: `/controlnet/txt2img` is the same route as `/sdapi/v1/txt2img`, and `/controlnet/img2img` is the same route as `/sdapi/v1/img2img`, except for an extra property that has been added at the root of the json object in the body: `"control_units": []`, which expects a list of `ControlNetUnitRequest`.

Each field of the `ControlNetUnitRequest` object has its own value and maybe we should look into clarifying what the expected value is for each. You can find each field at http://localhost/docs -> `model-ControlNetUnitRequest` at the bottom of the page.

You can start by specifying only the `"model": "..."` property IIRC, which expects one of the values returned by GET `/controlnet/model_list`. The defaults of all values are dumped at http://localhost/docs.

Here's an example img2img object that I used when testing the multi-controlnet api pr: (ellipsis `...` is for omitted base64 image values)

```json
{
  "init_images": [...],
  "resize_mode": 0,
  "denoising_strength": 0.5,
  "image_cfg_scale": 0,
  "mask_blur": 4,
  "inpainting_fill": 0,
  "inpaint_full_res": true,
  "inpaint_full_res_padding": 0,
  "inpainting_mask_invert": 0,
  "initial_noise_multiplier": 0,
  "prompt": "1girl",
  "styles": [],
  "seed": 351924614,
  "subseed": -1,
  "subseed_strength": 0,
  "seed_resize_from_h": -1,
  "seed_resize_from_w": -1,
  "sampler_name": "Euler",
  "batch_size": 1,
  "n_iter": 1,
  "steps": 50,
  "cfg_scale": 7,
  "width": 512,
  "height": 512,
  "restore_faces": false,
  "tiling": false,
  "negative_prompt": "",
  "eta": 0,
  "s_churn": 0,
  "s_tmax": 0,
  "s_tmin": 0,
  "s_noise": 1,
  "override_settings": {},
  "sampler_index": "Euler",
  "include_init_images": false,
  "controlnet_units": [
    {
      "model": "diff_control_sd15_depth_fp16 [978ef0a1]"
    }
  ]
}
```

In this case, the controlnet unit specified inherits the image specified in the `"init_images"` property. To specify a custom image, use the controlnet property `"input_image"`.

You can probably omit most of the root values. In my case I used a diff model. The api currently expects the `[hash]` to be part of the name, but we may want to also support only the name of the file without extension in the near future.

marcsyp:
This is great info, I think I understand better already -- only thing is that I'm not seeing the ControlNetUnitRequest documented in the localhost docs, so I can't see the fields available.  I only see 3 things documented related to controlnet: 

- GET
[/controlnet/model_list](http://127.0.0.1:7860/docs#/default/model_list_controlnet_model_list_get)
Model List

- POST
[/controlnet/detect](http://127.0.0.1:7860/docs#/default/detect_controlnet_detect_post)
Detect

And in Schemas:
<html>
<body>
<div id="model-Body_detect_controlnet_detect_post" class="model-container" data-name="Body_detect_controlnet_detect_post" style="box-sizing: border-box; background: rgba(0, 0, 0, 0.05); border-radius: 4px; margin: 0px 20px 15px; position: relative; transition: all 0.5s ease 0s;"><span class="model-box" style="background: none; border-radius: 4px; display: inline-block; padding: 10px;"><div class="model-box" style="box-sizing: border-box; background: none; border-radius: 4px; display: inline-block; padding: 10px;"><span class="model" style="color: rgb(59, 65, 81); font-family: monospace; font-size: 12px; font-weight: 600;"><span class=""><button aria-expanded="true" class="model-box-control" style="color: inherit; font: inherit; font-feature-settings: inherit; font-kerning: inherit; font-optical-sizing: inherit; font-palette: inherit; font-synthesis: inherit; font-variation-settings: inherit; forced-color-adjust: inherit; text-orientation: inherit; text-rendering: inherit; -webkit-font-smoothing: inherit; -webkit-locale: inherit; -webkit-text-orientation: inherit; -webkit-writing-mode: inherit; writing-mode: inherit; zoom: inherit; accent-color: inherit; place-content: inherit; place-items: inherit; place-self: inherit; alignment-baseline: inherit; animation: inherit; app-region: inherit; appearance: inherit; aspect-ratio: inherit; backdrop-filter: inherit; backface-visibility: inherit; background: inherit; background-blend-mode: inherit; baseline-shift: inherit; block-size: inherit; border-block: inherit; border-bottom: 0px; border-radius: inherit; border-collapse: inherit; border-end-end-radius: inherit; border-end-start-radius: inherit; border-image: inherit; border-inline: inherit; border-left: inherit; border-right: inherit; border-start-end-radius: inherit; border-start-start-radius: inherit; border-top: inherit; inset: inherit; box-shadow: inherit; box-sizing: inherit; break-after: inherit; break-before: inherit; break-inside: inherit; buffered-rendering: inherit; caption-side: inherit; caret-color: inherit; clear: inherit; clip: inherit; clip-path: inherit; clip-rule: inherit; color-interpolation: inherit; color-interpolation-filters: inherit; color-rendering: inherit; color-scheme: inherit; columns: inherit; column-fill: inherit; gap: inherit; column-rule: inherit; column-span: inherit; contain: inherit; contain-intrinsic-block-size: inherit; contain-intrinsic-size: inherit; contain-intrinsic-inline-size: inherit; container: inherit; content: inherit; content-visibility: inherit; counter-increment: inherit; counter-reset: inherit; counter-set: inherit; cursor: inherit; cx: inherit; cy: inherit; d: inherit; display: inherit; dominant-baseline: inherit; empty-cells: inherit; fill: inherit; fill-opacity: inherit; fill-rule: inherit; filter: inherit; flex: 1 1 0%; flex-flow: inherit; float: inherit; flood-color: inherit; flood-opacity: inherit; grid: inherit; grid-area: inherit; height: inherit; hyphenate-character: inherit; hyphenate-limit-chars: inherit; hyphens: inherit; image-orientation: inherit; image-rendering: inherit; initial-letter: inherit; inline-size: inherit; inset-block: inherit; inset-inline: inherit; isolation: inherit; letter-spacing: inherit; lighting-color: inherit; line-break: inherit; list-style: inherit; margin-block: inherit; margin: inherit; margin-inline: inherit; marker: inherit; mask: inherit; mask-type: inherit; math-depth: inherit; math-shift: inherit; math-style: inherit; max-block-size: inherit; max-height: inherit; max-inline-size: inherit; max-width: inherit; min-block-size: inherit; min-height: inherit; min-inline-size: inherit; min-width: inherit; mix-blend-mode: inherit; object-fit: inherit; object-position: inherit; object-view-box: inherit; offset: inherit; opacity: inherit; order: inherit; orphans: inherit; outline: inherit; outline-offset: inherit; overflow-anchor: inherit; overflow-clip-margin: inherit; overflow-wrap: inherit; overflow: inherit; overscroll-behavior-block: inherit; overscroll-behavior-inline: inherit; overscroll-behavior: inherit; padding-block: inherit; padding: 0px; padding-inline: inherit; page: inherit; page-orientation: inherit; paint-order: inherit; perspective: inherit; perspective-origin: inherit; pointer-events: inherit; position: inherit; quotes: inherit; r: inherit; resize: inherit; rotate: inherit; ruby-position: inherit; rx: inherit; ry: inherit; scale: inherit; scroll-behavior: inherit; scroll-margin-block: inherit; scroll-margin: inherit; scroll-margin-inline: inherit; scroll-padding-block: inherit; scroll-padding: inherit; scroll-padding-inline: inherit; scroll-snap-align: inherit; scroll-snap-stop: inherit; scroll-snap-type: inherit; scrollbar-gutter: inherit; shape-image-threshold: inherit; shape-margin: inherit; shape-outside: inherit; shape-rendering: inherit; size: inherit; speak: inherit; stop-color: inherit; stop-opacity: inherit; stroke: inherit; stroke-dasharray: inherit; stroke-dashoffset: inherit; stroke-linecap: inherit; stroke-linejoin: inherit; stroke-miterlimit: inherit; stroke-opacity: inherit; stroke-width: inherit; tab-size: inherit; table-layout: inherit; text-align: inherit; text-align-last: inherit; text-anchor: inherit; text-combine-upright: inherit; text-decoration: inherit; text-decoration-skip-ink: inherit; text-emphasis: inherit; text-emphasis-position: inherit; text-indent: inherit; text-overflow: inherit; text-shadow: inherit; text-size-adjust: inherit; text-transform: inherit; text-underline-offset: inherit; text-underline-position: inherit; touch-action: inherit; transform: inherit; transform-box: inherit; transform-origin: inherit; transform-style: inherit; transition: inherit; translate: inherit; user-select: inherit; vector-effect: inherit; vertical-align: inherit; visibility: inherit; border-spacing: inherit; -webkit-box-align: inherit; -webkit-box-decoration-break: inherit; -webkit-box-direction: inherit; -webkit-box-flex: inherit; -webkit-box-ordinal-group: inherit; -webkit-box-orient: inherit; -webkit-box-pack: inherit; -webkit-box-reflect: inherit; -webkit-highlight: inherit; -webkit-line-break: inherit; -webkit-line-clamp: inherit; -webkit-mask-box-image: inherit; -webkit-mask: inherit; -webkit-mask-composite: inherit; -webkit-print-color-adjust: inherit; -webkit-rtl-ordering: inherit; -webkit-ruby-position: inherit; -webkit-tap-highlight-color: inherit; -webkit-text-combine: inherit; -webkit-text-decorations-in-effect: inherit; -webkit-text-fill-color: inherit; -webkit-text-security: inherit; -webkit-text-stroke: inherit; -webkit-user-drag: inherit; -webkit-user-modify: inherit; white-space: inherit; widows: inherit; width: inherit; will-change: inherit; word-break: inherit; word-spacing: inherit; x: inherit; y: inherit; z-index: inherit;"><span class="pointer" style="cursor: pointer;"><span class="model-title" style="position: relative; color: rgb(80, 80, 80); font-family: sans-serif; font-size: 16px;"><span class="model-title__text">Body_detect_controlnet_detect_post</span></span></span><span class="model-toggle" style="cursor: pointer; display: inline-block; font-size: 10px; margin: auto 0.3em; position: relative; top: 6px; transform: rotate(90deg); transform-origin: 50% 50%; transition: transform 0.15s ease-in 0s;"></span></button><span class="brace-open object">{</span><span class="inner-object">

controlnet_module | Controlnet Module[...]
-- | --
controlnet_input_images | Controlnet Input Images[...]
controlnet_processor_res | Controlnet Processor Resolution[...]
controlnet_threshold_a | Controlnet Threshold a[...]
controlnet_threshold_b | Controlnet Threshold b[...]
 

</span><span class="brace-close">}</span></span></span></div></span></div><div id="model-Body_generate_classes_dreambooth_classifiers_post" class="model-container" data-name="Body_generate_classes_dreambooth_classifiers_post" style="box-sizing: border-box; background: rgba(0, 0, 0, 0.05); border-radius: 4px; margin: 0px 20px 15px; position: relative; transition: all 0.5s ease 0s;"><span class="models-jump-to-path" style="opacity: 0.65; position: absolute; right: 5px; top: 8px; color: rgb(59, 65, 81); font-family: sans-serif; font-size: medium; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;"></span><span class="model-box" style="background: none; border-radius: 4px; display: inline-block; padding: 10px; color: rgb(59, 65, 81); font-family: sans-serif; font-size: medium; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;"><br class="Apple-interchange-newline">
</body>
</html>

 I may be missing something about how the docs are used.

ljleb:
@marcsyp Do you see any `/sdapi/v1` routes? There should also be these routes in the docs:
- POST `/controlnet/txt2img`
- POST `/controlnet/img2img`

Plus the `ControlNetUnitRequest` schema.

If you don't see any `/sdapi/v1` routes, please refer to the discussion in #421. Might be related.

marcsyp:
I don't, which is weird -- I don't see any txt2img or img2img routes at all....  in the past I did see those routes, that's how I learned how to do my first successful POSTs on txt2img and img2img...

ljleb:
Yeah I'm not sure why it happens to be honest. One fix could be to hijack the webui `*2img` routes differently, so that they still appear even when the webui does not show basic api routes.

ljleb:
Note that even with a fix, soon the webui will start supporting always on scripts in api calls: https://github.com/AUTOMATIC1111/stable-diffusion-webui/issues/8253

When/if this PR is eventually merged, we will probably deprepcate/remove our own `/txt2img` and `/img2img` routes.

marcsyp:
so is this a bug in CN or in A1111?  How do I get back to seeing all the routes in the api docs, even just for basic a1111?  wait and keep trying?  :)

ljleb:
If you try removing controlnet you should still not see any `/sdapi/v1` route. It is probably not a controlnet issue. Maybe it is because you have encryption/middleware enabled, or something like that?

paulo-coronado:
Guys, I am experiencing the same issue as @kft334:

> "Can anyone confirm that Text-To-Image works? I'm getting bad outputs that seem like normal Text-To-Image. Everything loads and there are no errors though."

I am using the route `/sdapi/v1/txt2img` and the following payload:
<details>
<summary>Click to see the payload</summary>

```JSON
{
  "enable_hr": false,
  "denoising_strength": 0,
  "prompt": "...",
  "styles": [
    "string"
  ],
  "seed": -1,
  "subseed": -1,
  "subseed_strength": 0,
  "seed_resize_from_h": -1,
  "seed_resize_from_w": -1,
  "sampler_name": "DDIM",
  "batch_size": 1,
  "n_iter": 1,
  "steps": 20,
  "cfg_scale": 7,
  "width": 512,
  "height": 512,
  "restore_faces": false,
  "tiling": false,
  "negative_prompt": "...",
  "eta": 0,
  "s_churn": 0,
  "s_tmax": 0,
  "s_tmin": 0,
  "s_noise": 1,
  "override_settings": {},
  "script_args": [],
  "sampler_index": "DDIM",
  "controlnet_units": [
    {
      "module": "canny",
      "model": "control_canny-fp16 [e3fe7712]",
      "weight": 1,
      "resize_mode": "Envelope (Outer Fit)",
      "lowvram": false,
      "guessmode": false,
      "input_image": "..."
    }
  ]
}
```

</details>

Reading the thread, I tried the following options (**without success**):
- Tried DDIM and other samplers;
- Disabled installed scripts. I just kept the built-in ones (LDSR, Lora, ScuNET, SwinIR and prompt-bracket-checker);
- Enabled "Allow other script to control this extension".

No error message is shown.

Could you guys please help me? @sangww @ljleb @Mikubill @chrisbward

kiriri:
@paulo-coronado 
The route is /controlnet/txt2img

paulo-coronado:
My mistake @kiriri! I am using `/controlnet/txt2img` now and although the request is successful, the response generates an image that is not base64... I am using online [B64 > PNG converter](https://codebeautify.org/base64-to-image-converter) and the image is not formatted correctly (despite it looking like b64). No error message is shown.

alexcode0:
Hi all, thanks for the development and discussion in this thread, it's very helpful.
I've managed to get the controlnet working through the API, and I'm successfully generating a canny preprocess output.  How can I take this canny preprocess and use it to drive the canny model with txt2img via the API?
Thanks again

ljleb:
@alexcode0 you can pass preprocessed images as a base64 string to the `"input_image"` property of the controlnet processing unit you want to use. Then you put `"module": "none"` or just leave out this property, and use the `"model"` property to select the right canny model. Check the wiki for more info: https://github.com/Mikubill/sd-webui-controlnet/wiki/API

alexcode0:
> @alexcode0 you can pass preprocessed images as a base64 string to the `"input_image"` property of the controlnet processing unit you want to use. Then you put `"module": "none"` or just leave out this property, and use the `"model"` property to select the right canny model. Check the wiki for more info: https://github.com/Mikubill/sd-webui-controlnet/wiki/API#controlnetunitrequest-json-object
> 
> You will also find an example of the json structure via `localhost:7869/docs`.

Hm.  I tried exactly that before posting my Q.  Seemed like the logical way to implement it, but no luck.

`def run_sd():
    img = cv2.imread('canny.png')
    png_img = cv2.imencode('.png', img)
    preproc_64 = base64.b64encode(png_img[1]).decode('utf-8')

    #choose model for each layer
    option_payload = {
        "sd_model_checkpoint": "galaxytimemachinesGTM_v3.safetensors [f8ad2aafb5]",
        "CLIP_stop_at_last_layers": 2
    }
    response = requests.post(url=f'{url}/sdapi/v1/options', json=option_payload)
    payload = {
        "prompt": "A PURPLE VEST",
        "negative_prompt": "",
        "width": 512,
        "height": 512,
        "steps": 100,
        "cfg": 10,
        "sampler_index": "DPM++ 2S a Karras",
        "controlnet_units": [
            {
                "input_image": preproc_64,
                "mask": '',
                "module": "none",
                "model": "control_sd15_canny [fef5e48e]",
                "weight": 1.6,
                "lowvram": False,
                "processor_res": 512,
                "threshold_a": 64,
                "threshold_b": 64,
                "guidance": 1,
                "guidance_start": 0,
                "guidance_end": 1,
                "guessmode": True
            }
        ]
    }
    generate_image(payload,"output")`

`def generate_image(payload, filename):
    response = requests.post(url=f'{url}/controlnet/txt2img', json=payload)

    r = response.json()
    #print(r)
    for i in r['images']:
        image = Image.open(io.BytesIO(base64.b64decode(i.split(",", 1)[0])))

        png_payload = {
            "image": "data:image/png;base64," + i
        }
        response2 = requests.post(url=f'{url}/sdapi/v1/png-info', json=png_payload)

        pnginfo = PngImagePlugin.PngInfo()
        pnginfo.add_text("parameters", response2.json().get("info"))
        image.save(f'{filename}.png', pnginfo=pnginfo)
    return`

ljleb:
> Hm. I tried exactly that before posting my Q. Seemed like the logical way to implement it, but no luck.

Can you elaborate on what you expect to see, and then what you see instead happening? I don't have enough info at the moment to help.

alexcode0:
> > Hm. I tried exactly that before posting my Q. Seemed like the logical way to implement it, but no luck.
> 
> Can you elaborate on what you expect to see, and then what you see instead happening? I don't have enough info at the moment to help.

Essentially when I use the webui, I can upload an image to the control net, and get a preprocessed image, which is then ported to the txt2img workflow and is generated using the prompt and the preprocessed image as a guide.

When I try to do this via the API, I get the canny preprocess successfully.  When I pass that preprocessed image to the control net again and try to apply the model (control_sd15_canny [fef5e48e]), the output image is unchanged from the preprocessed input.  See image below:

Untitled

ljleb:
Thanks for the info. The api should return 2 images in your case, 1 for the preprocessor (unchanged when module=none) and 1 for the txt2img output. Are you sure only 1 image is returned?

As you save both images under the same name, I suspect the txt2img output is just overwritten by the preprocessor image.

ljleb:
You can check "don't save detectmap to output" in webui settings if you dont want the controlnet maps as output (should probably be an api parameter)

alexcode0:
> Thanks for the info. The api should return 2 images in your case, 1 for the preprocessor (unchanged when `"module": "none"`) and 1 for the txt2img output. Are you sure only 1 image is returned?
> 
> As you save both images under the same name, I suspect the txt2img output is just overwritten by the preprocessor image. Consider adding `i` to your filename somewhere, i.e. `f'{filename}-{i}'` and use `for i, img in enumerate(r['images'])` to iterate over your images

Bingo!  I thought that may have been the case and checked the response json but didn't find 2.  Went back and checked and yup, it's in there alright.

Thank you for taking the time to help

ImranBug:
`import requests
import base64

# Open the control image file in binary mode
with open("poses\poses_base.png", "rb") as f:
    # Read the image data
    image_data = f.read()

    # Encode the image data as base64
    image_base64 = base64.b64encode(image_data)

    # Convert the base64 bytes to string
    image_string = image_base64.decode("utf-8")

# Define the payload with the prompt and other parameters
payload = {
    "prompt": "A PURPLE VEST",
    "negative_prompt": "",
    "width": 512,
    "height": 512,
    "steps": 100,
    "cfg": 10,
    "sampler_index": "DPM++ 2S a Karras",
    "controlnet_units": [
        {
            "input_image": image_string,
            "mask": '',
            "module": "none",
            "model": "control_sd15_canny [fef5e48e]",
            "weight": 1.6,
            "lowvram": False,
            "processor_res": 512,
            "threshold_a": 64,
            "threshold_b": 64,
            "guidance": 1,
            "guidance_start": 0,
            "guidance_end": 1,
            "guessmode": True
        }
    ]
}
# Send the request to the API endpoint
response = requests.post(url=f'http://127.0.0.1:7860/controlnet/txt2img', json=payload)

# Print the response
print(response)
`

The response is giving 500 server error, can any one help me what am I doing wrong?

ljleb:
With latest version of webui it is possible that the controlnet API does not work anymore. There are efforts currently to put it back on its feet in #527

dbokser:
> # Send the request to the API endpoint
> response = requests.post(url=f'http://127.0.0.1:7860/controlnet/txt2img', json=payload)
> 
> # Print the response
> print(response) `
> 
> The response is giving 500 server error, can any one help me what am I doing wrong?

Have you tried decoding the response?

decoded = response.read().decode()

I also had a 500 Server Error but when decoding it gave me more information, telling me the input image wasn't encoded correctly.

But is there any update on whether controlnet API is working in the latest release? I tried both the /controlnet/txt2img and /sdapi/v1/txt2img endpoints and they both aren't going through controlnet

ljleb:
Waiting on approval for #587. The API will work again once it is merged into `main`.

Bruce-shuai:
> I don't, which is weird -- I don't see any txt2img or img2img routes at all.... in the past I did see those routes, that's how I learned how to do my first successful POSTs on txt2img and img2img...

hi, i also meet this problem, and you solve it?

PR #208, opened 2023-02-19
Webcam support

missionfloyd:
I see a lot of people on reddit using puppets for poses, so I figured this would be handy,

One small problem though, it flips the displayed image when you click the capture button. This is apparently a gradio [bug](https://github.com/gradio-app/gradio/issues/767). It's supposedly been [fixed](https://github.com/gradio-app/gradio/pull/774).

It doesn't seem to affect output images.


Mikubill:
Seems like too many checkboxes in the toolbar..but anyway, good idea!

missionfloyd:
It is a bit crowded. I considered putting them up by the image input. Or is there a better spot? Perhaps in settings?

missionfloyd:
How about buttons instead?
Untitled

PR #235, opened 2023-02-20
Added if (__name__ == '__main__'): 

stassius:
Added if (__name__ == '__main__'):  to the controlnet.py so it can be called from other scripts. It's needed for the further development of the api.

### DIFF
--- a/scripts/controlnet.py
+++ b/scripts/controlnet.py
@@ -614,8 +614,8 @@ def on_ui_settings():
 
     # control_net_skip_hires
 
-
-script_callbacks.on_ui_settings(on_ui_settings)
+if (__name__ == '__main__'):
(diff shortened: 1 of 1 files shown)


stassius:
@Mikubill In some of the last commits, this line was removed again and it broke my API build. How can I make sure it stays there? The problem is without it you can't import the module, it hangs.

stassius:
I think I got it. I shouldn't have added it to script_callbacks.on_ui_settings(on_ui_settings). Now it shows the Settings option again and works in my api. I'll make another PR.

ljleb:
Is the value of `__name__` ever going be be `'__main__'`? If that is not the case, maybe we should consider making a different file specifically for the purpose of importing from other extensions?

PR #248, opened 2023-02-20
Add 'hiresfix' and 'guidance_strength' for api

KotoriKoi:
If the api is called, Controlnet does not take effect. Please try to put api.py under sd-webui/scripts.


scripts/api.py
```diff
@@ -98,7 +98,12 @@ async def txt2img(
         controlnet_processor_res: int = Body(512, title='Controlnet Processor Res'),
         controlnet_threshold_a: int = Body(64, title='Controlnet Threshold a'),
         controlnet_threshold_b: int = Body(64, title='Controlnet Threshold b'),
-        controlnet_guidance: float = Body(1.0, title='ControlNet Guidance Strength'),
+        guidance_strength:float = Body(1.0, title='Guidance Strength'),
```
Mikubill, line 101: Guidance strength may be confused with other options/extensions, how about ControlNet Guidance Strength?

Issue #237, opened 2023-02-20
Add Automatic Prompt


cjfghk5697:
Hi, I have a question about the automatic prompt. In the paper, the authors implemented **automatic prompt(BLIP)**, and works well. Therefore, how about adding automatic prompt? The code of automatic prompt and ControlNet is in my [Colab](https://colab.research.google.com/drive/1arOOr1DAL3R2YRpDhy2E8yRf_89T-cJE?usp=sharing). Additionally, I tested the difference using the default Colab prompt and the automatic prompt. [Result](https://colab.research.google.com/drive/10CzdIwpRgS88JsNBWxzy3UwSkYY-upRv?usp=sharing)
The automatic prompt makes users convenient.

**In the paper :** 
> (3) Automatic prompt: In order to test the state-of-the-art maximized quality of a fully automatic
pipeline, we also try to use automatic image captioning methods (e.g., BLIP [34]) to generate prompts
using the results obtained by “default prompt” mode. We use the generated prompt to diffusion again.

![image](https://user-images.githubusercontent.com/80466735/220124858-572a3831-e480-4098-9ef6-9357c4565f2c.png)

demirklvc:
yes please!

cjfghk5697:
@Mikubill I weite code Automatic Prompt now. However, I can't find the code about the prompt. Can I ask the where is the Prompt code part?

Mikubill:
Prompt injection could be done in script.process I think:
https://github.com/Mikubill/sd-webui-controlnet/blob/2b370d9fde70acf72cc701bc113bbb20521db23b/scripts/controlnet.py#L402-L415
use `p(StableDiffusionProcessingTxt2Img).prompt` for injection or other stuff
ref: https://github.com/AUTOMATIC1111/stable-diffusion-webui/blob/0cc0ee1bcb4c24a8c9715f66cede06601bfc00c8/modules/txt2img.py#LL15C9-L15C41

cjfghk5697:
@Mikubill 
Thank u!! I open the PR.

PR #257, opened 2023-02-21
Add Automatic Prompt

cjfghk5697:
[Add Automatic Prompt #237](https://github.com/Mikubill/sd-webui-controlnet/issues/237)

I have been writing code about **Automatic Prompt**. However, it is not yet perfect now. Probably, this code is got problems. I will test this code this weekend.

### DIFF
--- a/example/test_api.ipynb
+++ b/example/test_api.ipynb
@@ -65,18 +65,18 @@
  ],
  "metadata": {
   "kernelspec": {
-   "display_name": "webui",
+   "display_name": "Python 3",
    "language": "python",
    "name": "python3"
   },
   "language_info": {
    "name": "python",
-   "version": "3.10.8"
+   "version": "3.10.4"
   },
   "orig_nbformat": 4,
   "vscode": {
    "interpreter": {
-    "hash": "bfdc572d05de84b0e25b3cefadad3fe196a31650b582a0267180110287606d68"
+    "hash": "3ad933181bd8a04b432d3370b9dc3b0662ad032c4dfaa4e4f1596c548f763858"
    }
   }
  },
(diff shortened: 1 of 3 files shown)


cjfghk5697:
- [ ] Add Option
```python
if p.prompt is "" and p.automatic:
   p.prompt=automatic_prompt(input_image,h,w)
```
If user want to use the Automatic Prompt, p.automatic option must be True.
- [ ] Add description

Mikubill:
Looks good! Maybe we should add some offload for lowvram users

cjfghk5697:
@Mikubill  
I agree that! Added to the checkbox.

cjfghk5697:
The BLIP model is cloning from the [github](https://github.com/salesforce/BLIP). You know that, this model is heavy. Therefore, you said, it is need to offload. 

I'm not sure, only when automatic is True, git clone and download the model.
```python
import git
git.Git("/").clone("https://github.com/salesforce/BLIP.git")
from BLIP.models.blip import blip_decoder

model_url = 'https://storage.googleapis.com/sfr-vision-language-research/BLIP/models/model_base_capfilt_large.pth'
```

Mikubill:
Hmm..maybe we could add some hints to tell user install BILP manually

cjfghk5697:
Oh... That's right and better solution😅I'll turn it into code before I fix it

cjfghk5697:
image
@Mikubill 
Hi! I tested it by adding code to add a ui once, but the checkbox is not added. How add the checkbox?  
```python
automatic = gr.Checkbox(label="Automatic Mode", value=False)
```

I'm really sorry to bother you

PR #262, opened 2023-02-21
refactor: multi-controlnet support

Mikubill:
~~Note that this may break current API and script control.~~ Fixed


Mikubill:
Quick test, also slightly improved memory usage.

| Source A | Source B | Output |
|:-------------------------:|:-------------------------:|:-------------------------:|
|  |   |  |

GeekyGhost:
This is looking good. Great work!

ClashSAN:
you will eventually get a user request for the slider to allow more than 5 controlnets, due to their re-usage of the same ones. 10 seems good enough.

henbucuoshanghai:
multi-controlnet。 where is the code?tks

PR #265, opened 2023-02-21
Add movie2movie

mattyamonaca:
Thanks for the great work.
I think it would be useful to have the function to generate videos from videos.
 I'll post an example below.

https://user-images.githubusercontent.com/48423148/220407218-7671fb98-77c9-4694-8c3b-973caff9a2f7.mp4

If necessary, I would be happy to have it merged after minor corrections are made.


AbyszOne:
COOL!!

OedoSoldier:
I think it'd better be an independent extension.

araimono:
Great job!

mattyamonaca:
Feature additions are largely complete.

Multi ControlNet adaptation has complicated the dependencies between codes.
I created Issue #335 to solve this.

If this issue is resolved, the extension will only be an independent file, making it easy to detach.

mattyamonaca:
adapted to multi controlnet

https://user-images.githubusercontent.com/48423148/221083184-e30c681e-0977-4aad-a059-a964e514a5a7.mp4

huizhen:
Great job. but how can we control the smooth interpolation and continuoutility between each frame's image like real video did.

AbyszOne:
It would be very important to know how it interacts with Img2Img.
1 - Can you use a reference image to condition the video?
2 - Could the same video be used in Img2Img simultaneously to retain much more coherence? Either in a batch of images or implemented within the extension itself.

mattyamonaca:
>but how can we control the smooth interpolation and continuoutility between each frame's image like real video did.

Currently only supports output in GIF format, needs to be expanded.
A parameter to set the time to output a single image has been added.

>It would be very important to know how it interacts with Img2Img.
>1 - Can you use a reference image to condition the video?
>2 - Could the same video be used in Img2Img simultaneously to retain much more coherence? Either in a batch of images or implemented within the extension itself.

I think both 1 and 2 are possible.
Basically, they are just connecting the images output by Control Net.

p0mad:
> I'll post an example below.

@Mikubill Is it possible to make the character fixed and stable( dont change person in the animation)?
I have created a 3Seconds video Pose (openpose +hands + face) to input to the SD+CN to get a single character/Avatar playing something ( like dance, ...) according to the input pose!

Is that possible? can you please point me into that direction?


P.S. [something like this](https://www.reddit.com/r/StableDiffusion/comments/14yljav/my_quest_for_consistent_animation_update/)

Thanks
Best regards

Issue #254, opened 2023-02-21
[Bug] After calling the api, all pictures affected by the controlnet


Oyaxira:
After I call controlnet/txt2img and pass the initial image,
when i request sdapi/v1/txt2img, the generated picture is obviously affected by the last picture sent to controlnet/txt2img

I try to read the source code of webui api and this project api but can't understand.
Judging from the log, the parameters passed into the interface are correct.
script_args is also an empty array in the processing flow of sdapi/v1/txt2img

Mikubill:
Could you share some log/reproduce steps/screenshots?

Oyaxira:
![result](https://user-images.githubusercontent.com/18203308/220322170-037981f4-169a-41b1-877a-2a27dd2ab43f.png)

source
1.jpg
2.jpg

Mikubill:
Fixed in https://github.com/Mikubill/sd-webui-controlnet/pull/278

Oyaxira:
The problem has been solved, thank you very much!

Issue #261, opened 2023-02-21
Change API thresholds to floats


kft334:
Thresholds don't work well with mlsd and normals currently. They should be changed to floats.

Also, returned control images should probably be in the correct format for reprocessing. At least mlsd and canny are both black foreground on white background atm. And a new endpoint should be added to return the model [hash] list.

PR #278, opened 2023-02-22
fixed api behavior

sangww:
summary of changes

1. fixed incorrect controlnet args added in later changes added in
2. fixed minor typos in the code that seemed fine functionally
3. removed boilerplate script args which seemed incompatible with other settings
4. added network restore routine in controlnet.py which is needed for the api to correctly function after initiating a controlnet generation.


Mikubill:
Looks good! It should fix https://github.com/Mikubill/sd-webui-controlnet/issues/254.
btw, also checkout https://github.com/Mikubill/sd-webui-controlnet/issues/261

sangww:
> Looks good! It should fix #254. btw, also checkout #261

Thanks for the note! Changed the thresholds to float.

PR #305, opened 2023-02-22
Feature request: write input image hash to metadata

JimAllanson:
This would calculate a hash for the input image, and write it out to the text metadata.

This would be especially useful to me, as I'm working on indexing images by the metadata encoded into the PNG files, and having an indicator of the original image used would allow grouping of output by common input image.

### DIFF
--- a/scripts/controlnet.py
+++ b/scripts/controlnet.py
@@ -1,6 +1,7 @@
 import gc
 import os
 import stat
+import hashlib
 from collections import OrderedDict
 
 import torch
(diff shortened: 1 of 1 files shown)


missionfloyd:
There's actually a [function](https://github.com/AUTOMATIC1111/stable-diffusion-webui/blob/0cc0ee1bcb4c24a8c9715f66cede06601bfc00c8/modules/hashes.py#L38-L46) you can use to simplify this.

```python
from modules.hashes import calculate_sha256

hash = calculate_sha256(args[4]["image"]) if args[4] else None
```

lllyasviel:
not sure abou this but will consider in the future versions

Issue #239, opened 2023-02-20
Error in generation


AndreyRGW:
```
line 555, in postprocess
    is_img2img_batch_tab = is_img2img and img2img_tab_tracker.submit_img2img_tab == 'img2img_batch_tab'
NameError: name 'img2img_tab_tracker' is not defined
```

Mikubill:
fixed in the latest commit

Issue #123, opened 2023-02-17
Error with using Batch of img2img


EkaLinjas:
![image](https://user-images.githubusercontent.com/125619279/219521316-426bb422-f31f-4db2-8e70-2b04b18ce164.png)
After finishing the first image of batch, an error occured and stop the process in here.
Is the extension hasn't support the Batch function in img2img?

in case of bad demonstrate here is the part of UI i use during this issue
![image](https://user-images.githubusercontent.com/125619279/219522186-25ae289f-53fc-4569-9b36-c76e66177a72.png)

Mikubill:
In settings tab, tick the "Do not append map to output" checkbox.

PR #316, opened 2023-02-23
Added checking the __name__, fixed issue with the settings entry

stassius:
I added the check for the script to be able to work with API. The issue of the settings menu entry not being shown is fixed.

### DIFF
--- a/scripts/controlnet.py
+++ b/scripts/controlnet.py
@@ -718,7 +718,8 @@ class Img2ImgTabTracker:
             return
 
 
-img2img_tab_tracker = Img2ImgTabTracker()
 script_callbacks.on_ui_settings(on_ui_settings)
-script_callbacks.on_after_component(img2img_tab_tracker.on_after_component_callback)
+if (__name__ == '__main__'):
+    img2img_tab_tracker = Img2ImgTabTracker()
+    script_callbacks.on_after_component(img2img_tab_tracker.on_after_component_callback)
 


Mikubill:
Looks good but will it breaks https://github.com/Mikubill/sd-webui-controlnet/issues/239?

stassius:
I didn't experience this issue, but you're right. The only line that breaks API was the last one. Fixed.

Mikubill:
What's the issue when batch detector exposed in main? Maybe we should fix it

ljleb:
I'm kind of confused because the value of `__name__` is never going to be `'__main__'`, since this is a webui extension. So the callback will simply never be registered with this change.

If the API needs to import code from the main script, maybe we should consider putting this code in a different location in the repo? I would like to know the conditions to reproduce this and debug if possible, so I can look for a fix along with you guys.

stassius:
> I'm kind of confused because the value of `__name__` is never going to be `'__main__'`, since this is a webui extension. So the callback will simply never be registered with this change (and so we might eventually see new reports similar to #123 IIUC).
> 
> If the API code needs to import/use code from the main script, maybe we should consider putting this code in a different location in the repo? I don't understand 100% the API code as I haven't got a lot of time to look into new changes as of recently. I would like to know the conditions to reproduce this and debug if possible, so I can look for a fix along with you guys.

Import the module from another script:
`importlib.import_module('extensions.sd-webui-controlnet.scripts.controlnet')`
Result: Automatic hangs on loading, Python is slowly consuming memory.

ljleb:
Thanks for the info.

> This PR solves it.
But it creates issues for webui use-cases that do not rely on the API, namely the callback will never be registered so batch tab detection will stop working. We could create a function that imports the script code using an internal mechanism to test whether the ui is importing the code or an external scripts is, and then redirect people to call this function. Also maybe there's already a way to write it as a test condition the way you are trying to do it, I'll try to look into it.

ljleb:
@stassius Can you try with `if __name__ == 'controlnet.py'` instead?

ljleb:
I allowed myself to update this PR with the proposed fix. Feel free to revert my changes if it does not work on your side @stassius.

ljleb:
As some people have had problems using the `__file__` global for some reason (see #446), maybe hardcoding `'controlnet.py'` will be a better idea.

Mikubill:
After recent updates, you can use `external code API` to interact with this extension (see https://github.com/Mikubill/sd-webui-controlnet/wiki/API). I will close this PR, feel free to reopen it if not works.

PR #350, opened 2023-02-24
fix: calculate size to fit output image

ZhUyU1997:
It is incorrect to scale to fit image. I add more calculations to fix it

|   Input Image  | Output Size    |  Output: AS-IS |  Output: TO-BE |
| :-----| ----: | :----: | :----: |
|  192x264 download | 192x264 |  | 
| 192x264 download | 264x192 |   | |


ZhUyU1997:
It looks like some issue was reported on #430 , If that, I prefer to discard this change.

PR #370, opened 2023-02-25
don't skip img2img when ControlNet is disabled

ddPn08:
Fixed `Skip img2img processing when using img2img initial image` skipping img2img even when ControlNet is disabled.


Mikubill:
enable variable not always exists - use `len(control_group)>0` instead

PR #374, opened 2023-02-25
add bug-report issue template

ljleb:
I replicated the bug-report issue template from the webui repo and updated/removed some sections. Is it okay to do it like this with respect to the license of the webui repo?

Not 100% sure if I adapted the issue template to something that matches this repo's needs. Feel free to mention anything that needs change.


Mikubill:
Looks good. I think it's enough for general issues.

Issue #399, opened 2023-02-27
Feature Request: Save preprocessor settings in PNG metadata


marcsyp:
It would be great to have the preprocessor settings saved in the PNG metadata because they can have a VERY big impact on final image result.  The preprocessor resolution and the two slider settings would be basically all that is needed to recreate an image.  Often I find myself needing to reload the UI and I have to write down these values or else I might lose my progress and have to start over with finding the perfect preprocessing settings.

Thanks!

marcsyp:
Note related Issue https://github.com/Mikubill/sd-webui-controlnet/issues/412#issue-1601653478

Jonseed:
This is one of the reasons I haven't really adopted ControlNet yet. There is no way to record how you generated an image, unless you do it by hand. None of the metadata of which image or its settings is stored with the generated images, so regenerating them later is a difficult task.

I really like [Ateist's idea](https://www.reddit.com/r/StableDiffusion/comments/118qock/comment/j9jbfqj/?utm_source=share&utm_medium=web2x&context=3) of adopting the format of LoRAs in the prompt itself to store that metadata, something like <controlnet:openpose="filename", guidance:0.2>.

The only additional thing would be the images themselves used as input, but I guess those could be stored in a kind of "library" folder somewhere.

ljleb:
At some point we could consider using the existing ui to generate a string to append to the prompt instead of driving the controlnet settings directly. This would make it possible both to use the UI and reproduce previous generations. To do this while preserving the angle bracket strings in the prompt, we can hijack `get_multicond_learned_conditioning` or `get_learned_conditioning` (to also consider the negative prompt) both defined in `modules.prompt_parser`.

To store pictures in metadata, I have seen one or two ideas being thrown around, namely one of them is indexing image hashes in a database and using hashes to reference them. Maybe a proximity hash could make it possible to use the next image hash if a hash cannot be found. If anyone knows an extension that already does something like this, it could be useful to link it here.

PR #407, opened 2023-02-27
Added support for Multi-ControlNet (list notation).

aiton-sd:
### Example:
- any = [any, any, any, ...]
- [any] = [any, None, None, ...]
- [None, None, any] = [None, None, any]
- [,,any] = [None, None, any]
- any, [,any,] = [any, any, any, ...], [None, any, None]

> Enabled Only:

- any = [any] = [any, None, None, ...]

### Note:

- Whitespace characters are replaced with None.
- The configuration values that correspond to the None index are used from GradioUI.
- There is a difference in behavior between 'Enabled' items and non-'Enabled' items when using 'any' (not in a list).
- Using [] in the filename will cause it to malfunction. Only the [hash value] is allowed.

### DIFF
--- a/scripts/xyz_grid_support.py
+++ b/scripts/xyz_grid_support.py
@@ -135,7 +135,7 @@ def add_axis_options(xyz_grid):
                     valslist[i] = sub_list + [None] * (max_length-len(sub_list))
 
         if not any(search_bracket(s) for s in valslist):  # There is no list inside
-            type_convert(valslist, type_func)  # Type conv
+            type_convert(valslist, type_func, allow_blank=True)  # Type conv
             return
         else:                            # There is a list inside
             fix(valslist, type_func)  # Fix & Type conv
--- a/scripts/xyz_grid_support.py
+++ b/scripts/xyz_grid_support.py
@@ -20,14 +20,30 @@ def debug_info(func):
     return debug_info_
 
 
-def find_xyz_grid():
+def find_dict(dict_list, keyword, search_key="name"):
+    return next((d for d in dict_list if d[search_key] == keyword), None)
+
+
+def flatten_list(lst):
+    for element in lst:
+        if isinstance(element, list):
+            yield from flatten_list(element)
+        else:
+            yield element
+
+
+def find_module(module_names):
+    if isinstance(module_names, str):
+        module_names = [s.strip() for s in module_names.split(",")]
+
     for data in scripts.scripts_data:
-        if data.script_class.__module__ in ["xyz_grid.py", "xy_grid.py"] and hasattr(data, "module"):
+        if data.script_class.__module__ in module_names and hasattr(data, "module"):
             return data.module
 
     return None
 
 
+# This function performs the main process of the module.
 def add_axis_options(xyz_grid):
     # This class is currently meaningless.
     class AxisOption(xyz_grid.AxisOption):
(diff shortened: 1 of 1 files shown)


Mikubill:
Looks good! Could you add some examples about how to apply different x/y/z to multi-controlnet configuration?

aiton-sd:
### Others Enabled
| XYZ Values             | Count | CN0         | CN1         | 
| ---------------------- | ----- | ----------- | ----------- | 
| None                   | 1     | UI          | UI          | 
| 2                      | 1     | 2           | 2           | 
| [2]                    | 1     | 2           | UI          | 
| [2, None]              | 1     | 2           | UI          | 
| [None, 2]              | 1     | UI          | 2           | 
| [, 2]                  | 1     | UI          | 2           | 
| [1, 2], [3, 4], [5, 6] | 3     | 1 / 3 / 5   | 2 / 4 / 6   | 
| None, [2], [, 2]       | 3     | UI / 2 / UI | UI / UI / 2 | 
### Enabled Only
| XYZ Values                | Count | CN0        | CN1           | 
| ------------------------- | ----- | ---------- | ------------- | 
| True                      | 1     | True       | UI            | 
| [True]                    | 1     | True       | UI            | 
| [None, 1], [True, true]   | 2     | UI / True  | True / True   | 
| [None, 0], [False, false] | 2     | UI / False | False / False |

marcsyp:
This is great -- I _think_ I understand the examples, but I'm a bit confused about what "Others Enabled" and "Enabled Only" mean.  Can you elaborate?  Also, what happens if I have CN-0 and CN-2 enabled, do I still provide an array of two values, or do I need to add a blank (e.g., [0,,2])?  Thanks!

marcsyp:
So I have been testing and there appear to be bugs -- following the [1, 2], [3, 4], [5, 6] pattern but with preprocessor resolution, I use:

[500, 500], [505, 505], [510, 510], [515, 515], [520, 520], [525, 525]

I do indeed get 7 images, but the last 5 images are nearly identical to image 2, which can be confirmed not to be the case when simply using the UI to set resolutions.

Also, I would expect there to be a syntax similar to standard syntax to handle ranges.  For instance, I have two preprocessers, and I want to test multiple resolutions between a range.  I might do:

[400-500, 600-700][11]  // results in 11 images at [400, 600], [410, 610], [420, 620]...
[400-500, ][11]  // results in 11 images at [400, UI], [410, UI], [420, UI]...
[400-500, 700-600][11]  // results in 11 images at [400, 700], [410, 690], [420, 680]...
[400-500, 600][11]  // results in 11 images at [400, 600], [410, 600], [420, 600]...

Also, there could be a way to add combinatorial ranges, where every possibility is created.  This could be done by simple leaving out the outer brackets and using range brackets for individual controlnets and leverages the standard syntax from webui:

600-700[11], 200-250(+10)  // results in a 11 x 6 grid with [600, 610, ... , 700] in X and [200, 210, ... , 250] in Y

Thoughts?

marcsyp:
My first thought after writing that is that the last example I gave allows for n-dimensional grids , which may not be a great idea.  :)  Maybe it only supports up to 3 ranges, and can't be used with other Y and Z controls in the X/Y/Z script...

marcsyp:
One additional thought -- debugging this is actually fairly difficult, because there is no preprocessor resolution or thresholds stored in the PNG data, so it is difficult to see how the X/Y/Z script is actually interpreting the values.  Referencing the following issue:

https://github.com/Mikubill/sd-webui-controlnet/issues/399#issue-1600272521

aiton-sd:
> This is great -- I _think_ I understand the examples, but I'm a bit confused about what "Others Enabled" and "Enabled Only" mean. Can you elaborate? Also, what happens if I have CN-0 and CN-2 enabled, do I still provide an array of two values, or do I need to add a blank (e.g., [0,,2])? Thanks!

If you input 'any' for items other than [ControlNet] Enabled, it will be applied to all ControlNets with Enabled checked on the UI. This is the same as inputting [any, any, any, ...].

If you input 'any' for [ControlNet] Enabled, it will override the Enabled value of Control Model - 0. This is the same as inputting [any].

So, if you want to apply Weight changes to CN0 and CN2 which are enabled on the UI, you should input 'any' or [any, , any]. If you want to control the Enabled of CN0 and CN2 with X/Y/Z plot, you need to input [any, , any].

marcsyp:
Sorry, I'm having a really hard time understanding -- what do 'any' and 'None' refer to exactly?  I'm not sure if you are saying that my examples above were not correct, or... ??  Maybe it's worth you posting some real-world examples of how the syntax is used, with demostrated use cases with the UI and actual X/Y/Z commands that result in successful grids?  I guess the more I read the comments on this thread the less I think I understand...  I'm beginning to think I must be missing some key knowledge that others are more familiar with, or maybe I am trying to use this for a purpose it was not intended for.  Thank you!

BugMonkey42335:
I think I have managed to figure this system out. We're comparing are two variables on a single ui axis and you have to manually enter every possible combination. Good luck to anyone trying to compare more than two values.

marcsyp:
Yes but even this doesn't work properly, see [comment above](https://github.com/Mikubill/sd-webui-controlnet/pull/407#issuecomment-1451133014).

marcsyp:
P.S. I really appreciate all the work that everyone is putting into this project -- just trying to help the development on what will be a really awesome and critical feature for fine-tuning parameters.  Thanks all.

marcsyp:
OK, apologies, a correction -- the notation for multiple passes is indeed working as described above -- my test was for very small increments in preprocessor resolution that do not have a large impact on the image result, and so I was incorrectly assuming that the list processing had failed.  Further test with larger increments do in fact work properly.

Regardless, my comments above regarding the processing of a syntax to automatically generate lists are still valid.  It would be great to be able to define an explicit number of pairings using something like:

 [400-500, 600-700][11] // results in 11 images at [400, 600], [410, 610], [420, 620]...
[400-500, ][11] // results in 11 images at [400, UI], [410, UI], [420, UI]...
[400-500, 700-600][11] // results in 11 images at [400, 700], [410, 690], [420, 680]...
[400-500, 600][11] // results in 11 images at [400, 600], [410, 600], [420, 600]...

In the meantime, I will simply be writing code to generate the text pairs spelled out the way I want them and copy paste those values into the field.  Thanks.

PR #336, opened 2023-02-24
GL: Set CanvasWidth/Height from input image size; tooltips, send output size 

GeorgLegato:
Resolves #378 

# Video about latest button-rework 
(smaller, tooltips, near to the action)

https://user-images.githubusercontent.com/7210708/222928094-f3978652-233f-4bdb-896f-9afdf931fd73.mp4

# Proportional-Scale-Panel
image

# ToolTips! and "Send to Render size" implemented
image

# Quick-buttons to setup 512/768 proportional scaling, then click on "send to output size"
image

---
already merged these fixes from this branch:

Resolves #114 
image
image
image

Issue #311, opened 2023-02-23
[f] what about putting preview annotation and input image in the same row?


lllyasviel:
right now the preview annotation looks a little bit big - not very handy to scroll forward and backward

Mikubill:
fixed in https://github.com/Mikubill/sd-webui-controlnet/commit/b0c6d973380eb8fdd2d53387ebce4071cb1e8e5b.

PR #483, opened 2023-03-04
Image cropping

### LINKED ISSUE #449: Feature Request, Cropping on CN image input
Would it be possible to include the same cropping function in img2img for CN inputs?

image

thank you

missionfloyd:
Adds an option to crop input images. Closes #449

The image `tool` parameter apparently can't be updated, so I had to make a new image control (`cropped_image`). Clicking the ✂ button will swap between `input_image` and `cropped_image`. Changes to `cropped_image` are automatically sent to `input_image`.

I did run into a problem where `cropped_image` wouldn't get resized to make room for annotator result, resulting in overlapping, so I moved it to the next row. Maybe there's a better way to deal with that. Tabs, perhaps?
Untitled
Untitled2


jonwong666:
bless you

jonwong666:
just tested, works great!

scripts/controlnet.py
```diff
@@ -218,12 +220,15 @@ def uigroup(self, is_img2img):
         infotext_fields = []
         with gr.Row():
             input_image = gr.Image(source='upload', mirror_webcam=False, type='numpy', tool='sketch')
+            cropped_image = gr.Image(label="Crop image", interactive=True, tool="select", visible=False)
+        with gr.Row():
```
Mikubill, line 224: This may break previous fix for https://github.com/Mikubill/sd-webui-controlnet/issues/311 - Is there any other way to make it smaller?

scripts/controlnet.py
```diff
@@ -62,9 +62,11 @@
 switch_values_symbol = '\U000021C5' # ⇅
 camera_symbol = '\U0001F4F7'        # 📷
 reverse_symbol = '\U000021C4'       # ⇄
+scissors_symbol = '\U00002702'      # ✂
 
 webcam_enabled = False
 webcam_mirrored = False
+crop_enabled = False
```
ljleb, line 69: Having this global means all units of both txt2img and img2img will have the same state. Consider instantiating an `gr.State` to pass around state to gradio component updates.

scripts/controlnet.py
missionfloyd: I made the annotator result the same height as the input. I don't know if it can fit on the same row as cropped. Maybe I can make it fit with CSS.

scripts/controlnet.py
missionfloyd: I didn't know about that, thanks.

scripts/controlnet.py
missionfloyd: How about tabs? This way, you don't need the "hide annotator" button either.
image

And I was thinking "preview annotator" could be a toolbutton next to preprocessor. Something like 🔍.

GeorgLegato: oh we are messing on buttons/icons. 
please check my pr #336 to see how i shrinked  buttons/rearranged. we would clash a bit here, if my pr gets merged

missionfloyd: I'll leave the buttona alone then.

jonwong666:
any chance this can be merged into the main repo please? I constantly use this feature, but everytime I update to the latest, it disappears. thank you

JonDadley:
I'd also like to see this merged into main, it's very useful

ljleb:
This looks clean, I'll test to see if the conversation above should be marked as resolved.

lllyasviel:
i am going to close this since ui has changed a lot, feel free to open another pr if this is still needed

PR #515, opened 2023-03-07
In this video I have explained how to install from scratch

FurkanGozukara:
I am pretty sure this video will be very helpful to new users. I recorded video today so it is extremely upto date and clear.

The video has video chapters and fully manually fixed subtitles.

I hope you consider to add this video.

Thank you so much for your efforts.


Mikubill:
Thank you for your detailed and useful videos. Since we want the readme to be as informative as possible, we may add your link to the wiki in the future.

PR #520, opened 2023-03-07
refactor: delete flatten_list, using itertools

Guillaume-Fgt:
A little refactor to simplify the xyz file. Instead of a function looping over a list to flatten it, I propose to use itertools.chain(). The code is shorter and faster.

Example:
```
import itertools
from time import perf_counter

my_list = [[1, 2, 3], [4, 5, 6], [7], [8, 9]]


def flatten_list(lst):
    result = []
    for element in lst:
        if isinstance(element, list):
            result.extend(flatten_list(element))
        else:
            result.append(element)
    return result


time_1 = perf_counter()
merged = list(itertools.chain(*my_list))
time_2 = perf_counter()

time_3 = perf_counter()
merged_2 = flatten_list(my_list)
time_4 = perf_counter()

print(merged)
print(time_2 - time_1)
print(merged_2)
print(time_4 - time_3)
```
output:

> [1, 2, 3, 4, 5, 6, 7, 8, 9]
> 2.200016751885414e-06
> [1, 2, 3, 4, 5, 6, 7, 8, 9]
> 3.200024366378784e-06

flatten_list was only used with string field like Resize Mode. So I tested it generating a x/y/z plot with values "Envelope (Outer Fit),Scale to Fit (Inner Fit)" and it worked like before.

I didn't find test to run in the code, so I don't know what other checks I can do.

### DIFF
--- a/scripts/xyz_grid_support.py
+++ b/scripts/xyz_grid_support.py
@@ -1,5 +1,6 @@
 import re
 import numpy as np
+import itertools
 
 from modules import scripts, shared
 
@@ -29,18 +30,9 @@ def find_dict(dict_list, keyword, search_key="name", stop=False):
         raise ValueError(f"Dictionary with value '{keyword}' in key '{search_key}' not found.")
 
 
-def flatten_list(lst):
-    result = []
-    for element in lst:
-        if isinstance(element, list):
-            result.extend(flatten_list(element))
-        else:
-            result.append(element)
-    return result
-
-
 def is_all_included(target_list, check_list, allow_blank=False, stop=False):
-    for element in flatten_list(target_list):
+    flatten_list = list(itertools.chain(target_list))
+    for element in flatten_list:
         if allow_blank and str(element) in ["None", ""]:
             continue
         elif element not in check_list:


scripts/xyz_grid_support.py
```diff
@@ -29,18 +30,9 @@ def find_dict(dict_list, keyword, search_key="name", stop=False):
...
-        if isinstance(element, list):
-            result.extend(flatten_list(element))
-        else:
-            result.append(element)
-    return result
-
-
 def is_all_included(target_list, check_list, allow_blank=False, stop=False):
-    for element in flatten_list(target_list):
+    flatten_list = list(itertools.chain(target_list))
```
Mikubill, line 34: Maybe we should use `*target_list` instead of `target_list`?

Guillaume-Fgt: This is what I thought at the beginning. But in fact, the xs parameter passed to this function is a list like

`["Envelope (Outer Fit)", "Scale to Fit (Inner Fit)"]`

So if you put a *, it will transform it like

> ['E', 'n', 'v', 'e', 'l', 'o', 'p', 'e', ' ', '(', 'O', 'u', 't', 'e', 'r', ' ', 'F', 'i', 't', ')', 'S', 'c', 'a', 'l', 'e', ' ', 't', 'o', ' ', 'F', 'i', 't', ' ', '(', 'I', 'n', 'n', 'e', 'r', 
> ' ', 'F', 'i', 't', ')']

Without, I have the same output as before:

> ['Envelope (Outer Fit)', 'Scale to Fit (Inner Fit)']

aiton-sd: The target of the function flatten_list is the list xs edited by the ListParser class.
The elements of xs may contain no list at all, for example:
`[1, 2, 3, 4, 5, 6, 7, 8, 9, 10]`
or may contain a mix of lists and non-list elements, for example:
`[1, 2, [3, 4], [5, 6, 7], 8, 9]`

itertools.chain() does not work properly for such lists.

Guillaume-Fgt: Hi aiton, could you explain me how the file work, I should have miss something. 

flatten_list() is called in only one place: is_all_included() function.
is_all_included() is only called in one place, confirm()
In confirm(), the condition to call is_all_included() is:
`isinstance(func_or_str, str)`

So if I check the possibilities to call it, I see:
```
AxisOption("[ControlNet] Model", identity, apply_field("control_net_model"), confirm=confirm("model"), choices=choices_model, cost=0.9),
 AxisOption("[ControlNet] Resize Mode", identity, apply_field("control_net_resize_mode"), confirm=confirm("resize_mode"), choices=choices_resize_mode),
 AxisOption("[ControlNet] Preprocessor", identity, apply_field("control_net_module"), confirm=confirm("preprocessor"), choices=choices_preprocessor),
```
For Model, it is the list of ControlNet models generated in controlnet.py

For resize_mode, the possible values are 

> ["Envelope (Outer Fit)", "Scale to Fit (Inner Fit)", "Just Resize"]

For preprocessor, possible values are:

>         self.preprocessor = {
>             "none": lambda x, *args, **kwargs: (x, True),
>             "canny": canny,
>             "depth": midas,
>             "depth_leres": leres,
>             "hed": hed,
>             "mlsd": mlsd,
>             "normal_map": midas_normal,
>             "openpose": openpose,
>             "openpose_hand": openpose_hand,
>             "clip_vision": clip,
>             "color": color,
>             "pidinet": pidinet,
>             "scribble": simple_scribble,
>             "fake_scribble": fake_scribble,
>             "segmentation": uniformer,
>             "binary": binary,
>         }

So I believed flatten_list was only dealing with strings... When is it called for values like [1, 2, [3, 4], [5, 6, 7], 8, 9] for example?

aiton-sd: The problem is that the list is irregular.
itertools.chain() does not work properly for irregular lists.

Assuming XYZ Values is the following string:
"canny, depth, [canny, depth], [canny]"

Applying the following process in xyz_grid.py:
`valslist = [x.strip() for x in chain.from_iterable(csv.reader(StringIO(vals)))] `

results in the following list, which is the argument xs for the function confirm_:
`['canny', 'depth', '[canny', 'depth]', '[canny]']`

This list is edited by the ListParser class to become:
`['canny', 'depth', ['canny', 'depth'], ['canny', None]]`

When flatten_list is applied to this list, it results in:
`['canny', 'depth', 'canny', 'depth', 'canny', None]`

However, applying list(itertools.chain(*my_list)) results in:
`['c', 'a', 'n', 'n', 'y', 'd', 'e', 'p', 't', 'h', 'canny', 'depth', 'canny', None]`

Guillaume-Fgt: ok, I understand your point. What I am saying is with the current code, this case is never met.

After testing the three cases: X/Y/Z plot for ControlNet Model, Resize_Mode and Preprocessor, `flatten_list = itertools.chain(target_list)` worked because the list it encountered were (values obtained by printing the list in `choices_model()`, `choices_resize_mode()` and `choices_preprocessor()`):
```
model_list = [
    "diff_control_sd15_canny_fp16 [ea6e3b9c]",
    "diff_control_sd15_depth_fp16 [978ef0a1]",
    "diff_control_sd15_hed_fp16 [86db5d7c]",
    "diff_control_sd15_mlsd_fp16 [14f0845a]",
    "diff_control_sd15_normal_fp16 [00173cc1]",
    "diff_control_sd15_openpose_fp16 [1723948e]",
    "diff_control_sd15_scribble_fp16 [1f29174d]",
    "diff_control_sd15_seg_fp16 [b9c1cc12]",
]

resize_list = ["Envelope (Outer Fit)", "Scale to Fit (Inner Fit)", "Just Resize"]

preprocessor_list = [
    "none",
    "canny",
    "depth",
    "depth_leres",
    "hed",
    "mlsd",
    "normal_map",
    "openpose",
    "openpose_hand",
    "clip_vision",
    "color",
    "pidinet",
    "scribble",
    "fake_scribble",
    "segmentation",
    "binary",
]
```

Maybe I am just missing a case somewhere, I don't know...

aiton-sd: The variable xs is a list generated based on the string input from XYZ Values.
xs is passed as the first argument to is_all_included function as follows:
`is_all_included(xs, check_list, allow_blank=True, stop=True)`

Then, the function is_all_included applies flatten_list to the first argument target_list = xs.
```
def is_all_included(target_list, check_list, allow_blank=False, stop=False):
    for element in flatten_list(target_list):
```

The list obtained by the function choices_foo is assigned to the second argument check_list.
`elif element not in check_list:`

Guillaume-Fgt: ok, I think I got it. In fact, it was a feature from x/y/z plot I was ignoring.

You can pass in a string field like preprocessor [canny', 'depth] and it will do a combination of it: canny, depth, canny-depth and canny-none... I was aware of this kind of thing for integer and floating fields. I thought it wasn't the case for string. 

Thank you for the explanations. Sorry for this bad PR...

PR #527, opened 2023-03-07
Deprecate `/controlnet/*2img` web API in favor of new alwayson_scripts support

### LINKED ISSUE #567: [API-UPDATE] Need to update to accommodate recent API changes in AUTOMATIC1111
# What Happened
- https://github.com/AUTOMATIC1111/stable-diffusion-webui/pull/8187 got merged recently.
- This allows user to use Controlnet and other scripts from the `/sdapi/v1/txt2img` and `sdapi/v1/img2img` with parameter ` alwayson_scripts`.
- This will allow us to now deprecate the `/controlnet/txt2img` and `/controlnet/img2img` routes as mentioned in wiki.

# Issue 1
- I haven't understood entire flow of the AUTOMATIC1111's API but I will try my best to explain this.
- Currently `ScriptRunner` class has `process` method which is responsible to pass on arguments to the other scripts like Controlnet.
- ScriptRunner.process will call Script.process method with expected method signature of `process(self, p, *args):` which will allow all scripts to be run with API.
- For Controlnet script's Script.process method has different signature ` process(self, p, is_img2img=False, is_ui=False, *args)` which is causing trouble if we want to just directly use the Controlnet from the  `/sdapi/v1/txt2img` API because first two parameters will go towards the `is_img2img` and `is_ui`.
- So current schema for the Controlnet parameters will become as following for solution of this issue:
```python
"ControlNet": [
  false,  # is_img2img
  false,  # is_ui
  true,   # enabled,
  "canny",    # module,
  "control_canny-fp16 [e3fe7712]",    # model,
  1,    # weight,
  {"image":"BASE64_IMAGE_STRING"},    # image,
  false,    # scribble_mode,
  "Scale to Fit (Inner Fit)",   # resize_mode,
  false,    # rgbbgr_mode,
  false,    # lowvram,
  512,    # pres,
  100,    # pthr_a,
  200,    # pthr_b,
  0,    # guidance_start,
  1,    # guidance_end,
  false   # guess_mode
]
```
Ref: https://github.com/Mikubill/sd-webui-controlnet/blob/5bf7e54e8ef82f013d4b304184b60cdb52810362/scripts/controlnet.py#L610
This is current workaround and I am mentioning this because Second issue is dependent on this.

Curl Command for the Request :https://pastebin.com/432EG3PT

# Issue 2
- If you go ahead and send the curl request mentioned above then you will encounter KeyError for 'image' for following line of code: https://github.com/Mikubill/sd-webui-controlnet/blob/5bf7e54e8ef82f013d4b304184b60cdb52810362/scripts/controlnet.py#L751
- Basically image['image'] is currently string not numpy array of image.
- This is also mentioned in the description of the https://github.com/AUTOMATIC1111/stable-diffusion-webui/pull/8187
- >For exemple, controlnet might have to detect that the images are a base64 string instead of an Image but it shouldn't require much work.
- I will be raising PR for this detection check. But till then it won't work. Edit: Raised PR(#568)

I am creating this issue so we can keep track of the API related changes. Also let me know if Issue 1 needs to be changing but that is something you should handle as you have more knowledge of the API flow.

# Tasks
- [x] Raise PR for image str detection(#568)
- [ ] Mark APIs `/controlnet/txt2img` and `/controlnet/img2img`  as deprecated.
- [ ] Update wiki accordingly. [I guess this needs to be done by owner]

### LINKED ISSUE #571: [Bug]: I can't get API call to load any Controlnet process

### What happened?

I trying to run this extension via API call using /controlnet/txt2img
The txt2img run fine but the controlnet model doesn't kick in. 
No log in console that the model was loaded. The result image didn't look like it has any controlnet applied.

Using it via UI works fine thou.

This is my POST body. I have stripped out the base64 image, model and the prompts.
```json
{
    "enable_hr": false,
    "hr_scale": 1,
    "hr_upscaler": "Latent (nearest-exact)",
    "hr_second_pass_steps": 20,
    "hr_resize_x": 0,
    "hr_resize_y": 0,
    "denoising_strength": 0.4,
    "firstphase_width": 0,
    "firstphase_height": 0,
    "prompt": "prompt...",
    "styles": [],
    "seed": 1841357043,
    "subseed": -1,
    "subseed_strength": 0.0,
    "seed_resize_from_h": -1,
    "seed_resize_from_w": -1,
    "batch_size": 1,
    "n_iter": 1,
    "steps": 25,
    "cfg_scale": 10,
    "width": 512,
    "height": 512,
    "restore_faces": false,
    "tiling": false,
    "negative_prompt": "negative prompt...",
    "eta": 0,
    "s_churn": 0,
    "s_tmax": 0,
    "s_tmin": 0,
    "s_noise": 1,
    "override_settings": {
        "sd_model_checkpoint": "model...",
        "CLIP_stop_at_last_layers": 2
    },
    "sampler_name": "Euler a",
    "sampler_index": "Euler a",
    "script_name": null,
    "script_args": [],
    "controlnet_units": [
        {
            "input_image": "base64img...",
            "mask": "",
            "module": "none",
            "model": "difference_controlSd15Openpose [1723948e]",
            "weight": 1.0,
            "lowvram": false,
            "processor_res": 64,
            "threshold_a": 64,
            "threshold_b": 64,
            "guidance": 1.0,
            "guidance_start": 0.0,
            "guidance_end": 1.0,
            "guessmode": false
        }
    ]
}
```

1. Call API /controlnet/txt2img with provided JSON body
2. Wait for result.

It should have generates image with controlnet applied as works with UI

webui: 
AUTOMATIC1111/stable-diffusion-webui@ab0b0e1e767ddc7ceeaced11513c3692538938eb
controlnet: 

Google Chrome

### Command Line Arguments

```Shell
--xformers --no-half-vae --listen --api --enable-insecure-extension-access
```

### Console logs

```Shell
Nothing weird
```

### Additional information

_No response_

ljleb:
Closes #571, closes #567

This combines the 2 argument passing strategies we use in `processing` and `postprocessing` into a single strategy. (using an array of structures) I find that code readability increases with this.

As I am not sure whether this really is a good idea, I created a companion discussion: #528. This changes a lot of code, risks breaking something as there are no tests. Feel free to close if it is not welcome.

### DIFF
--- a/scripts/api.py
+++ b/scripts/api.py
@@ -121,8 +121,12 @@ class ApiHijack(api.Api):
     def controlnet_any2img(self, any2img_request, original_callback, is_img2img):
         any2img_request = nest_deprecated_cn_fields(any2img_request)
         any2img_request.alwayson_scripts.update({'ControlNet': {'args': [is_img2img, False, *[to_api_cn_unit(unit) for unit in any2img_request.controlnet_units]]}})
+        controlnet_units = any2img_request.controlnet_units
         delattr(any2img_request, 'controlnet_units')
-        return original_callback(self, any2img_request)
+        result = original_callback(self, any2img_request)
+        result.parameters['controlnet_units'] = controlnet_units
+        result.parameters['alwayson_scripts'].clear()
+        return result
 
 api.Api = ApiHijack
 
--- a/scripts/api.py
+++ b/scripts/api.py
@@ -120,7 +120,7 @@ class ApiHijack(api.Api):
 
     def controlnet_any2img(self, any2img_request, original_callback, is_img2img):
         any2img_request = nest_deprecated_cn_fields(any2img_request)
-        any2img_request.alwayson_scripts.update('ControlNet', {'args': [is_img2img, False, *[vars(unit) for unit in any2img_request.controlnet_units]]})
+        any2img_request.alwayson_scripts.update('ControlNet', {'args': [is_img2img, False, *[to_api_cn_unit(unit) for unit in any2img_request.controlnet_units]]})
         delattr(any2img_request, 'controlnet_units')
         return original_callback(self, any2img_request)
 
--- a/scripts/api.py
+++ b/scripts/api.py
@@ -4,18 +4,13 @@ import numpy as np
 from fastapi import FastAPI, Body
 from PIL import Image
 import copy
-import contextlib
 import pydantic
 import sys
 
 import gradio as gr
 
-from modules import ui
 from modules.api.models import *
 from modules.api import api
-from modules.processing import StableDiffusionProcessingTxt2Img, StableDiffusionProcessingImg2Img
-
-import modules.scripts as scripts
 
 from scripts import external_code
 from scripts.processor import *
@@ -119,6 +114,7 @@ class ApiHijack(api.Api):
         )
 
     def controlnet_any2img(self, any2img_request, original_callback, is_img2img):
+        warn_deprecated_route(is_img2img)
         any2img_request = nest_deprecated_cn_fields(any2img_request)
         alwayson_scripts = dict(any2img_request.alwayson_scripts)
         any2img_request.alwayson_scripts.update({'ControlNet': {'args': [is_img2img, False, *[to_api_cn_unit(unit) for unit in any2img_request.controlnet_units]]}})
@@ -142,7 +138,6 @@ def nest_deprecated_cn_fields(any2img_request):
     if all(v is None for v in deprecated_cn_fields.values()):
         return any2img_request
 
-    warn_deprecated_cn_params()
     deprecated_cn_fields = {k[len(cn_root_field_prefix):]: v for k, v in deprecated_cn_fields.items()}
     for k, v in deprecated_cn_fields.items():
         if v is None:
@@ -178,10 +173,11 @@ def to_api_cn_unit(unit_request: ControlNetUnitRequest) -> external_code.Control
         guess_mode=unit_request.guessmode,
     )
 
-def warn_deprecated_cn_params():
+def warn_deprecated_route(is_img2img):
+    route = 'img2img' if is_img2img else 'txt2img'
     warning_prefix = '[ControlNet] warning: '
-    print(f"{warning_prefix}using deprecated '{cn_root_field_prefix}*' request params", file=sys.stderr)
-    print(f"{warning_prefix}consider using the 'control_units' request param instead", file=sys.stderr)
+    print(f"{warning_prefix}using deprecated '/controlnet/{route}' route", file=sys.stderr)
+    print(f"{warning_prefix}consider using the '/sdapi/v1/{route}' route with the 'alwayson_scripts' json property instead", file=sys.stderr)
 
 def controlnet_api(_: gr.Blocks, app: FastAPI):
     @app.get("/controlnet/model_list")
--- a/scripts/api.py
+++ b/scripts/api.py
@@ -108,8 +108,6 @@ class ApiHijack(api.Api):
         return self.controlnet_any2img(
             any2img_request=txt2img_request,
             original_callback=ApiHijack.text2imgapi,
-            p_class=StableDiffusionProcessingTxt2Img,
-            script_runner=scripts.scripts_txt2img,
             is_img2img=False,
         )
 
@@ -117,40 +115,16 @@ class ApiHijack(api.Api):
         return self.controlnet_any2img(
             any2img_request=img2img_request,
             original_callback=ApiHijack.img2imgapi,
-            p_class=StableDiffusionProcessingImg2Img,
-            script_runner=scripts.scripts_img2img,
             is_img2img=True,
         )
 
-    def controlnet_any2img(self, any2img_request, original_callback, p_class, script_runner, is_img2img):
+    def controlnet_any2img(self, any2img_request, original_callback, is_img2img):
         any2img_request = nest_deprecated_cn_fields(any2img_request)
-        script_runner = create_cn_script_runner(script_runner, any2img_request.controlnet_units, is_img2img)
+        any2img_request.alwayson_scripts.update('ControlNet', {'args': [is_img2img, False, *[vars(unit) for unit in any2img_request.controlnet_units]]})
         delattr(any2img_request, 'controlnet_units')
-        with self.queue_lock:
-            self_copy = copy.copy(self)
-            self_copy.queue_lock = contextlib.nullcontext()
-            with OverrideInit(p_class, scripts=script_runner):
-                return original_callback(self_copy, any2img_request)
+        return original_callback(self, any2img_request)
 
 api.Api = ApiHijack
-
-class OverrideInit:
-    def __init__(self, cls, **kwargs):
-        self.cls = cls
-        self.kwargs = kwargs
-        self.original_init = None
-
-    def __enter__(self):
-        def init_hijack(p, *args, **kwargs):
-            self.original_init(p, *args, **kwargs)
-            for k, v in self.kwargs.items():
-                setattr(p, k, v)
-
-        self.original_init = self.cls.__init__
-        self.cls.__init__ = init_hijack
-
-    def __exit__(self, exc_type, exc_val, exc_tb):
-        self.cls.__init__ = self.original_init
 
 def nest_deprecated_cn_fields(any2img_request):
     deprecated_cn_fields = {k: v for k, v in vars(any2img_request).items()
@@ -174,43 +148,6 @@ def nest_deprecated_cn_fields(any2img_request):
 
     any2img_request.controlnet_units.insert(0, ControlNetUnitRequest(**deprecated_cn_fields))
     return any2img_request
-
-def create_cn_script_runner(script_runner: scripts.ScriptRunner, control_unit_requests: List[ControlNetUnitRequest], is_img2img: bool):
-    if not script_runner.scripts:
-        script_runner.initialize_scripts(False)
-        ui.create_ui()
-
-    cn_script = external_code.find_cn_script(script_runner)
-    cn_script_runner = copy.copy(script_runner)
-    cn_script_runner.alwayson_scripts = [cn_script]
-    cn_script_args = [None] * cn_script.args_from
-    cn_units = [to_api_cn_unit(control_unit_request) for control_unit_request in control_unit_requests]
-    external_code.update_cn_script_in_place(
-        script_runner=cn_script_runner,
-        script_args=cn_script_args,
-        cn_units=cn_units,
-        is_img2img=is_img2img,
-    )
-
-    def make_script_runner_f_hijack(fixed_original_f):
-        def script_runner_f_hijack(p, *args, **kwargs):
-            original_script_args = p.script_args
-            try:
-                p.script_args = cn_script_args
-                fixed_original_f(p, *args, **kwargs)
-            finally:
-                p.script_args = original_script_args
-
-        return script_runner_f_hijack
-
-    for k in ('process', 'process_batch', 'postprocess', 'postprocess_batch', 'postprocess_image'):
-        original_f = getattr(cn_script_runner, k, None)
-        if original_f is None:
-            continue
-
-        setattr(cn_script_runner, k, make_script_runner_f_hijack(original_f))
-
-    return cn_script_runner
 
 def to_api_cn_unit(unit_request: ControlNetUnitRequest) -> external_code.ControlNetUnit:
     input_image = to_base64_nparray(unit_request.input_image) if unit_request.input_image else None
--- a/scripts/api.py
+++ b/scripts/api.py
@@ -120,7 +120,7 @@ class ApiHijack(api.Api):
 
     def controlnet_any2img(self, any2img_request, original_callback, is_img2img):
         any2img_request = nest_deprecated_cn_fields(any2img_request)
-        any2img_request.alwayson_scripts.update('ControlNet', {'args': [is_img2img, False, *[to_api_cn_unit(unit) for unit in any2img_request.controlnet_units]]})
+        any2img_request.alwayson_scripts.update({'ControlNet': {'args': [is_img2img, False, *[to_api_cn_unit(unit) for unit in any2img_request.controlnet_units]]}})
         delattr(any2img_request, 'controlnet_units')
         return original_callback(self, any2img_request)
 
--- a/scripts/api.py
+++ b/scripts/api.py
@@ -120,12 +120,13 @@ class ApiHijack(api.Api):
 
     def controlnet_any2img(self, any2img_request, original_callback, is_img2img):
         any2img_request = nest_deprecated_cn_fields(any2img_request)
+        alwayson_scripts = dict(any2img_request.alwayson_scripts)
         any2img_request.alwayson_scripts.update({'ControlNet': {'args': [is_img2img, False, *[to_api_cn_unit(unit) for unit in any2img_request.controlnet_units]]}})
         controlnet_units = any2img_request.controlnet_units
         delattr(any2img_request, 'controlnet_units')
         result = original_callback(self, any2img_request)
         result.parameters['controlnet_units'] = controlnet_units
-        result.parameters['alwayson_scripts'].clear()
+        result.parameters['alwayson_scripts'] = alwayson_scripts
         return result
 
 api.Api = ApiHijack
--- a/scripts/controlnet.py
+++ b/scripts/controlnet.py
@@ -3,7 +3,7 @@ import os
 import stat
 from collections import OrderedDict
 from enum import Enum
-from typing import Union
+from typing import Union, Dict, Any
 
 import torch
 
@@ -18,6 +18,7 @@ from scripts.processor import *
 from scripts.adapter import PlugableAdapter
 from scripts.utils import load_state_dict
 from scripts.hook import ControlParams, UnetHook
+from scripts import external_code
 from modules import sd_models
 from modules.paths import models_path
 from modules.processing import StableDiffusionProcessingImg2Img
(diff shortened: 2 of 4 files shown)


ljleb:
Yes of course, this is a work in progress. (hence still a draft) I haven't finished updating the code, just a rough idea of the change at the moment.

ljleb:
Maybe we should add tests to the repo before this? I have no prior professional experience with python, so I don't know how to set them up properly. I can open a PR for tests but it will probably be the wrong way to do them.

ljleb:
Quick test of 1 controlnet unit seems to work atm. Remains to be tested:
- [ ] multi controlnet
- [ ] gradio default values
- [ ] web api

ljleb:
For clarity, this PR makes it so that it's possible to call the web API `/sdapi/v1/*2img` routes like this:

```json
{
    ...
    "alwayson_scripts": {
        "ControlNet": {
            "args": [false, true,
            {   // 1
                "image": "base64...",
                "model": "model of choice..."
            }, {// 2
                "image": "base64...",
                "model": "second model of choice..."
            }]
        }
    }
}
```

PhoenixCreation:
Hi @ljleb, this is about the adding tests to the repo. This repo is actually quite big right now to start writing tests for each method. But we can add test cases for common end to end cases. But for that a lot of efforts will be required. 

I haven't looked at entire code flow but if we can agree on the several test cases then I can create one PR with them to get things started.

As you mentioned that paths from webui does not get resolved for independent test cases. For that If you can somehow share the code you have worked on then I might be able to help as I have good knowledge in testing with python.

If you want to create PR for test cases then it would be fine too. I might jump into that PR for suggestions if you think that's okay.

ljleb:
(apologies in advance for long comment...)

@PhoenixCreation Thanks for the help! Yes I think it is unrealistic to aim to test everything. Starting with something anywhere between 1 to 20 tests would really help, and then we could consider adding more test cases over time whenever we fix a bug, add a new feature or find a regression.

At the moment I am relying a lot on manually testing this PR. Having some tests would increase my confidence that this does not break anything. I think I would prefer for you to start an effort in this direction, if that's okay with you of course.

If you'd prefer that someone else start the base testing PR, then I could try putting something together real quick. I would really appreciate if you could provide some guidance into writing the tests and setting up the automatic discovery mechanism correctly. If you want, another option could be to co-author the PR?

As for what this current PR does:
- replaces the long list of arguments passed to `process` and `postprocess` with instances of [`ControlNetUnit`](https://github.com/Mikubill/sd-webui-controlnet/blob/4c13542c931f9401a4072b2f7fc9eb704e74a2f8/scripts/external_code.py#L13). This tidies up the args list into something easier to use and brings the cool perk of being able to use default values in the web API
- additionally, for backwards compatibility, the old way of calling `process` is still supported. To differentiate between a flattened list of controlnet args and `ControlNetUnit`, see [this function](https://github.com/Mikubill/sd-webui-controlnet/blob/fbb6354ba7b4394a2564afd976b7ab89b8cd5fa8/scripts/controlnet.py#L737). Basically, if `args[0]` of an arguments pack is of type `bool`, we assume it's a flattened tuple; otherwise, it has to be a `dict` or `ControlNetUnit`.

I am not totally certain of whether we really need to backwards support the flattened process args. I would think we should be able to just drop support for them, as developers are expected to use the external code API or web API to interface with the extension. Backwards compatibility is something we have to pay in code complexity, so I'm open to removing this feature. I'd say tests for this have a lower priority than the public facing API code + gradio interface. (if there's any way to test the gradio interface 😅)

For the web API, my intention with this PR is that the name of the properties do not have to be changed from the existing `controlnet_units` object properties in the deprecated API. (i.e. users can still use `guessmode` instead of `guess_mode`, check the current docs for property names that are expected to work)

So in other words, I believe we ideally need to verify that: (this is only for suggestive purposes, feel free to break it down into a different set of test cases)
- `process` and `postprocess` work as expected with `*args` of type `tuple[ControlNetUnit, ...]`
- main use cases of the web API with intended script arguments result in passing the right arguments to `process` and `postprocess`. This can be broken further into having to verify that:
    - default values are filled in when using dicts in API
    - explicit properties are passed down to `process`
- main use cases of [`external_code.update_cn_script_in_place`](https://github.com/Mikubill/sd-webui-controlnet/blob/4c13542c931f9401a4072b2f7fc9eb704e74a2f8/scripts/external_code.py#L138) and/or companion functions do not break. Basically, create a dummy `ControlNetUnit` and verify that changing properties result in calling `process` and `postprocess` with the right values
- for backwards compatibility (may not be needed?):
    - `process` and `postprocess` work with flattened list arguments `*args: tuple[Any, ...]`
    - extend the tests to the deprecated `/controlnet/*2img` routes

Again, it may be unrealistic to add all of these cases in the initial test PR. I tried to put them in order of my subjective perception of their priority.

For your suggestion on `is_img2img` and `is_ui`, I am not sure if I understand your proposition correctly. Do you mean i.e. to also allow to pass them as a dict for example? (and so instead of `bool, bool, *List[ControlNetUnit]`, you could also receive `dict[str, Any], *List[ControlNetUnit]`) I think the ideal implementation would be to not have to pass any of them to the webui by API. Both are there only for convenience in `process` and `postprocess`. I added them as I am not aware of a more reliable way to obtain this information. If you have a better idea or if I did not understand your suggestion, please let me know.

ljleb:
closes #571

ljleb:
I manually tested the main use cases: txt2img/img2img in web API, txt2img/img2img in gradio interface.

Marked as ready for review, but maybe we should consider waiting to have a couple of automated tests before merging.

PhoenixCreation:
> For your suggestion on is_img2img and is_ui, I am not sure if I understand your proposition correctly. Do you mean i.e. to also allow to pass them as a dict for example? (and so instead of bool, bool, *List[ControlNetUnit], you could also receive dict[str, Any], *List[ControlNetUnit]) I think the ideal implementation would be to not have to pass any of them to the webui by API. Both are there only for convenience in process and postprocess

I was asking to just get rid of them from API schema. It is not good design that you will have two booleans in array then object of actual information. So in simple words, just try to remove them.

PhoenixCreation:
@ljleb, And about testing, I will create separate PR for that and we can take discussion about test cases there. Keep this PR for API changes only.

As for list of test cases provided by you seems okay although difficulty to implement them will be the deciding factor to weather implement them or not. Also I would like others(specially @Mikubill) to also provide more insights about which things needs to be tested and which not.

Also I am planning to use pytest instead of unit_test(builtin) as it is more convenient for developers. Do let me know if we want to keep it to inbuilt modules.

Mikubill:
Looks good. I've added a basic txt2img api test (using unit_test tho), we can add/discuss more test cases later.
Thanks for contributions!

physis123:
I'm not sure if this is the right place but I'm now getting the following error when trying to call controlnet via API:

``` 
image = unit.get_image_dict()
AttributeError: 'ControlNetUnit' object has no attribute 'get_image_dict'
```

I'm using /sdapi/v1/img2img/ with this payload:
```
  init_images: [image],
      prompt: prompt,
      negative_prompt: negativePrompt,
      steps: options.steps,
      width: options.width,
      height: options.height,
      cfg_scale: options.cfg_scale,
      sampler_index: options.sampler_index,
      denoising_strength: options.denoising_strenth,
      alwayson_scripts: {
        "controlnet": {
          image: image,
          weight: 1,
          guidance: 1,
          guessmode: false,
          module: options.controlnet_module,
          model: options.controlnet_model,
      }
    },
```

I get back a normal img2img response but ControlNet isn't used. 

Am I doing something wrong?

PhoenixCreation:
Hi @Hugo-Matias, About the args being dictionary rather than list is something coming from AUTOMATIC 1111's API. That is something out of scope of this repo. Although schema of args list is currently bit tricky and I am working to see if we can make it something like 
```json
"args": [
       {
            "enabled": true,
            ....
       }, 
       { 
            "enabled": true,
             ... 
       }
]
```

Although, I also need clarification from @ljleb about how to pass images? currently I am using following schema:
```json
args: [false,false,
    {
    ...
    "input_image": {
            "image": "BASE64"
         }
     }
]
```

But this structure seems to be giving me following error:
`Traceback (most recent call last):
  File "C:\Tools\stable-diffusion-webui\modules\scripts.py", line 409, in process
  File "C:\Tools\stable-diffusion-webui\extensions\sd-webui-controlnet\scripts\controlnet.py", line 673, in process
    while len(image['mask'].shape) < 3:
KeyError: 'mask'`

From the code it seems that we now need to pass the mask compulsory?

aiton-sd:
@Mikubill Please revert the merge. There is a bug in the WebUI which is preventing the settings from being reflected.

Hugo-Matias:
> About the args being dictionary rather than list is something coming from AUTOMATIC 1111's API.

@PhoenixCreation I see, was afraid that might be the case. It's a very unfriendly approach if you need to pass the information around in object models.

I'm still unsure about the proper key strings for the CN unit, I was under the impression that `input_image` was just renamed to `image` but looks like its a dict now with both `image` and `mask`? I need to take a better look at the code but this PR should probably need a bit more work before merging.

PhoenixCreation:
>  I need to take a better look at the code but this PR should probably need a bit more work before merging.

@Hugo-Matias It is already merged but seems it has some issues as pointed out by @aiton-sd, I think we need to revert the changes of this MR and need to test it little bit more before we apply this changes.

physis123:
Just as an amateur user I would really appreciate if changes to the API were documented before being merged. I'm reading this thread trying to understand how to change the API call but I just keep getting different errors.

Hugo-Matias:
>  It is already merged but seems it has some issues as pointed out by @aiton-sd, I think we need to revert the changes of this MR and need to test it little bit more before we apply this changes.

Indeed, sorry that's what I meant with before merging and as @physis123 stated we should also focus on documentation parity.

ljleb:
> @PhoenixCreation  Although, I also need clarification from @ljleb about how to pass images? currently I am using following schema:

To pass images, you just pass the base64 string to `"input_image"` as before. To pass a mask, for now you have to pass literally `{"image": ..., "mask": ...}`. I didn't manually test this case, fixing in new commit. Basically the conversion from dict to dto needs to accept 2 different fields image and mask.

aiton-sd:
I apologize for the inconvenience.
The cause of the failure to reflect the settings may have been due to a malfunction of the Gradio server. Currently, it is being reflected successfully.
Timing was unfortunate. Sorry for the trouble.😭

ljleb:
@Hugo-Matias I get back a normal img2img response but ControlNet isn't used.

You need to call it like this:

```
{
    "alwayson_scripts": {
        "controlnet": {
            "args": [...]
        }
    }
}
```

Otherwise the webui will not pass any argument to the extension.

aiton-sd:
@Mikubill Thank you👍and I'm sorry.🙇

PhoenixCreation:
Hey @ljleb, It has been really confusing switching API schema. Would you be able to put one sample payload in PR description then it would help lot of people coming here unless we change the documentation.

ljleb:
I can update the wiki in fact. I'll add an additional section to show how to call `/sdapi/v1/*2img`. At the moment you can still call old `/controlnet/*2img` so i'll keep the section under it after merging.

ljleb:
@Hugo-Matias And, perhaps it's due to a poor implementation on Automatic's scripting system but in my opinion, is_img2img shouldnt be inferred from the endpoint call?

I really want to know how to do this. Maybe there is already this information available somewhere, but as the webui isn't documented I couldn't find a better way. Script classes have `is_img2img` and `is_txt2img` members but after testing id did not seem to change whether calling from img2img or txt2img. I'll test again maybe just to be sure, as this is surprising behavior to me.

By default they are set to false because the web API used to call the script without any arguments. all args needed default values. For now I did not attempt remove the default values yet because I was afraid to break existing code from dev users.

Hugo-Matias:
Unfortunately I can't be much help either, I've played a bit with the script example but never really explored it in depth and never applied an api to it. I'm just trying to think why that boolean is really needed when the endpoint itself is already conditioned to the value.

> By default is_ui and is_img2img pos args are set to false because the web API used to call the script without any arguments. all args needed default values. I believe this is no more necessary. For now I did not attempt remove the default values yet because I was afraid to break existing code from dev users.

I see, but as it is right now, passing is_ui as True would still call the webui as False right? Or is there another method that reads the proper argument value? Sorry for my lack of knowledge about the overall repo code.

ljleb:
> I see, but as it is right now, passing is_ui as True would still call the webui as False right? Or is there another method that reads the proper argument value? Sorry for my lack of knowledge about the overall repo code.

If you pass `[true, true, ...]` to the api the chsnges will be reflected inside `process`. Default values don't take effect when passed explicitly as argument. Ideally we should infer these values, but also allowing to override them means to be able to act as if api was a gui. Some use cases for this is when a script wants to pretend to call controlnet as if from gradio, even though the controlnet units actually originate from API.

Hugo-Matias:
> If you pass [true, true, ...] to the API the changes will be reflected inside process. ...

Ah ok, I think I was getting confused with the [controlnet_any2img](https://github.com/Mikubill/sd-webui-controlnet/blob/6a5580ca7c144e04220d1e12c1e0cc92eca8bccb/scripts/api.py#L116) method of the ApiHijack class, I see that's for routing the deprecated endpoints to standard relevant webui enpoint using the `alwayson_scripts` property.

Regarding the [ControlNetUnit object fields](https://github.com/Mikubill/sd-webui-controlnet/blob/6a5580ca7c144e04220d1e12c1e0cc92eca8bccb/scripts/api.py#L36), are the ones [included in the wiki](https://github.com/Mikubill/sd-webui-controlnet/wiki/API#controlnetunitrequest-json-object) still the valid ones apart from the `input_image` dict case? I think we are still missing a couple (`scribble_mode`, `rgbbgr_mode`) right?

Vespinian:
A bit of context from what I understand on the auto1111 api changes for those who haven't gone into the weeds. Feel free to correct me if I am mistaken.

Before there wasn't anyway to pass args to always on scripts (scripts that have always_visible == true like cnet) through the auto1111 *2img API. This repo worked around that limitation by cleverly hijacking the api and making their own routes. Normally the scriptrunner, which runs all scripts, has a giant list called script_args which, from what I could tell, apart from index 0, is fed by the gradio elements of the webui who just dump their values into that list at an assigned index. By hijacking the api and making their own scriptrunner, this repo was able to make a nice and friendly api schema which abstracted/fixed/bypassed this process.

The new alwayson_scripts param in the API makes it possible to pass the args to those scripts the same way the webui does, which is why it's a list and is a bit different from the current cnet api schema as it is the same args that are feed from the webui. It basically emulates what the webui would pass. Unfortunately, a description of what args to pass is not available unless the extension provides it through some means or you dive in the code, add a print and deduce it yourself. A change of the design of the scriptrunner in auto1111 could be done but you run the risk of breaking every extension that currently exist.

Hugo-Matias:
@Vespinian Do you know if it's possible to get the inference mode from the endpoint, ie. txt2img or img2img? Having a way to know that would help simplify the call arguments.

ljleb:
For clarity, I updated the wiki. Added a section on how to migrate, what the expected structure is, etc. Also added a banner at the top of the web API section that redirects to the migration section.

Vespinian:
@Hugo-Matias
You might be able to check which script_runner is being used in p.scripts either scripts.scripts_txt2img or scripts.scripts_img2img see [L219](https://github.com/AUTOMATIC1111/stable-diffusion-webui/blob/master/modules/api/api.py#L219), [L245](https://github.com/AUTOMATIC1111/stable-diffusion-webui/blob/master/modules/api/api.py#L245), [L271](https://github.com/AUTOMATIC1111/stable-diffusion-webui/blob/master/modules/api/api.py#L271), [L300](https://github.com/AUTOMATIC1111/stable-diffusion-webui/blob/master/modules/api/api.py#L300) in the api.py of auto1111

Hugo-Matias:
That's interesting, so we could potentially use that to initialize the `is_img2img` variable without passing it as an argument. Perhaps we could use a default value if the failsafe is triggered or just throw an exception...

Vespinian:
Well thinking about it a bit more, maybe we could type check the p object itself. It should be either a StableDiffusionProcessingTxt2Img or a StableDiffusionProcessingImg2Img.

ljleb:
Upon further testing, it seems that `script.txt2img` reflects the nature of the call in api. I'll remove the argument in the PR.

ljleb:
> Well thinking about it a bit more, maybe we could type check the p object itself. It should be either a StableDiffusionProcessingTxt2Img or a StableDiffusionProcessingImg2Img.

I tried this approach before, but we are swapping the pipeline to txt2img in some cases despite actually being in the img2img tab.

Vespinian:
> Upon further testing, it seems that `script.is_txt2img` and `script.is_img2img` reflects the nature of the call in api. I'll remove the argument in the PR.

Oh yeah, that makes sense, it's passed to [initialize_scripts](https://github.com/AUTOMATIC1111/stable-diffusion-webui/blob/master/modules/scripts.py#L280) and that passes it down to the scripts in the scriptrunner. You see it in action in the api.py when no scriptrunner is found [L221](https://github.com/AUTOMATIC1111/stable-diffusion-webui/blob/master/modules/api/api.py#L221) and [L273](https://github.com/AUTOMATIC1111/stable-diffusion-webui/blob/master/modules/api/api.py#L273)

Hugo-Matias:
> Upon further testing, it seems that `script.is_txt2img` and `script.is_img2img` reflects the nature of the call in api. I'll remove the argument in the PR.

Great news! Hopefully it will be enough to run things smoothly and without edge cases.

> One potential issue we may encounter by doing this is if users replace controlnet script with a copy in a script runner. I don't think this is a blocking issue as it will probably not happen, but it could lead to surprising behavior.

Would that situation happen with a regular api call on an unaltered SD installation? I don't entirely understand the issue and how it would affect the script's behavior but I can help with testing things once I'm done remapping objects and migrating my calls.

ljleb:
> Would that situation happen with a regular api call on an unaltered SD installation? I don't entirely understand the issue and how it would affect the script's behavior but I can help with testing things once I'm done remapping objects and migrating my calls.

Regular calls would not encounter this, but extensions or external code interfacing with scripts could do this. Not very likely to happen as you say, so it could be a valid solution IMO.

PR #543, opened 2023-03-08
xyz_refactor

Guillaume-Fgt:
- Remove the closure for the `confirm()` function and use `functools.partial()` instead.
- Separated the `add_axis_options()` in small pieces: no more nested function inside it.
- `extra_axis_options` is now a dict outside of any function, I think it is possible in the future to make it even simpler to avoid repeating the same pattern for each option.

@aiton-sd: your thoughts about it?

### DIFF
--- a/scripts/xyz_grid_support.py
+++ b/scripts/xyz_grid_support.py
@@ -1,3 +1,4 @@
+from functools import partial
 import re
 import numpy as np
 
@@ -18,6 +19,7 @@ def debug_info(func):
         if DEBUG_MODE:
             print(f"Debug info: {func.__name__}, {args}")
         return func(*args, **kwargs)
+
     return debug_info_
 
 
@@ -26,7 +28,9 @@ def find_dict(dict_list, keyword, search_key="name", stop=False):
     if result or not stop:
         return result
     else:
-        raise ValueError(f"Dictionary with value '{keyword}' in key '{search_key}' not found.")
+        raise ValueError(
+            f"Dictionary with value '{keyword}' in key '{search_key}' not found."
+        )
 
 
 def flatten(lst):
@@ -51,7 +55,7 @@ def is_all_included(target_list, check_list, allow_blank=False, stop=False):
     return True
 
 
-class ListParser():
+class ListParser:
     """This class restores a broken list caused by the following process
     in the xyz_grid module.
         -> valslist = [x.strip() for x in chain.from_iterable(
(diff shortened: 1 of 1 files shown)


aiton-sd:
> Remove the closure for the confirm() function and use functools.partial() instead

If it has the same structure as the function apply_field, it suggests that it is used for a similar purpose.
There is no need to break it down and use partial.


> extra_axis_options is now a dict outside of any function, I think it is possible in the future to make it even simpler to avoid repeating the same pattern for each option.

This method is efficient, but the structure of the list is visually confusing.
Additionally, in xyz_grid.py, axis_options is defined as follows:
```
axis_options = [
    AxisOption("Nothing", str, do_nothing, format_value=format_nothing),
    AxisOption("Seed", int, apply_field("seed")),
    .........,
]
```
I prioritized uniformity and structured extra_axis_options as a list with the same structure as axis_options.
However, I believe this is a matter of preference.
If you decide to make such a change, you should move the validation_data list closer to it.

I also think that the following changes would harm readability (although I understand the intention).
```
        raise ValueError(
            f"Dictionary with value '{keyword}' in key '{search_key}' not found."
        )
```

scripts/xyz_grid_support.py
```diff
@@ -302,127 +316,177 @@ def find_module(module_names):
...
+    {
+        "name": "preprocessor",
+        "type": str,
+        "check": choices_preprocessor,
+        "exclude": None,
+    },
+]
+
+
+def identity(x):
```
aiton-sd, line 394: Please restore the deleted comments as it is useful for understanding the process that this module is performing.

scripts/xyz_grid_support.py
```diff
@@ -302,127 +316,177 @@ def find_module(module_names):
...
+def apply_field(field):
+    @debug_info
+    def apply_field_(p, x, xs, *, field=field, enclosure=apply_field):
+        enable_script_control()
+        setattr(p, field, x)
+
+    return apply_field_
+
+
+def add_axis_options(xyz_grid, extra_axis_options: list[dict]) -> None:
```
aiton-sd, line 411: I think it's best to write functions in the order they will be used, and place lists that are frequently modified towards the bottom. For example, like this:
```
find_module
add_axis_options
identity
enable_script_control
apply_field
confirm
bool_
choices_for
make_excluded_list
validation_data
extra_axis_options
```

scripts/xyz_grid_support.py
```diff
@@ -302,127 +316,177 @@ def find_module(module_names):
...
+    return x
+
+
+def enable_script_control():
+    shared.opts.data["control_net_allow_script_control"] = True
+
+
+def apply_field(field):
+    @debug_info
+    def apply_field_(p, x, xs, *, field=field, enclosure=apply_field):
```
aiton-sd, line 404: Please make the following changes. I forgot to delete this.😅
`def apply_field_(p, x, xs):`

Guillaume-Fgt:
Thank you for your reply. I understand the will to stick with xyz_grid style of code. I believe nested functions and closures are more complicated to read, maintain and test but I agree it should stay that way.

So I restarted from the original xyz_grid_support and modified just 2 things:

- extended the comment `Starting the main process of this module.` with the order of execution of functions you provided above and reorganized the functions accordingly (basically moved `identity()` up)
- did the small modification on `apply_field()` you asked

scripts/xyz_grid_support.py
Guillaume-Fgt: done.

scripts/xyz_grid_support.py
Guillaume-Fgt: Added in comments for clarity for new coders. Good idea?

scripts/xyz_grid_support.py
Guillaume-Fgt: Restored. As it was just a draft, I deleted some parts here and there. Agree it's important.

Issue #549, opened 2023-03-09
Feature Request: Using the original generated txt2image as controlnet input


djkacevedo:
Ie. If I want a randomly generated bird in a specific cage, right now, I could generate the bird, move it to control net input, then turn on the first and second control nets, with the first being the bird and the second being the cage.

It would be nice if those extra steps could be automated, and I could just check a "use generated image," or something similar, and have the qebui just do an extra image gen at the start to get the control net input.

PR #557, opened 2023-03-10
Use generated im

Guillaume-Fgt:
In response to enhancement request Mikubill/sd-webui-controlnet#549.

I started to add the feature. In txt2img, if you enable ControlNet without image, you check the box "use generated image", you input a prompt and control net will automatically create an image and use it to preprocess/process.

Animation

What is done so far:
- creation of a checkbox in `uigroup()`
- add an if statement in `process()`

I have to set p.scripts() to None otherwise in `processing.process_images_inner()`, with the check on p.scripts, it kept to sent it back to controlnet.py and I was stuck in an infinite loop.
I do a check on the seed to avoid creating the generated image several times in the case of multicontrolnet.

@djkacevedo , did I understand your request correctly?

Part I am unsure / to do work:

- the cache of models, I don't know how it works for the moment
- the conditions with img2img, if there is an image or not. Not sure at all the way I added mine do not create a bug
- how do you actually mix 2 control net images in txt2img tab? I am aware of the method in img2img, playing with the denoising strenght, not in txt2img

### DIFF
--- a/scripts/controlnet.py
+++ b/scripts/controlnet.py
@@ -1,3 +1,4 @@
+from copy import deepcopy
 import gc
 import os
 import stat
@@ -69,7 +70,7 @@ tossup_symbol = '\u2934'
 webcam_enabled = False
 webcam_mirrored = False
 
-PARAM_COUNT = 15
+PARAM_COUNT = 16
 
 
 class ToolButton(gr.Button, gr.components.FormComponent):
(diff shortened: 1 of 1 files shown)


djkacevedo:
@Guillaume-Fgt Awesome! Yep. That's exactly what I was asking for.

PR #583, opened 2023-03-13
Revert "Deprecate `/controlnet/*2img` web API in favor of new alwayso…

Mikubill:
Reverts Mikubill/sd-webui-controlnet#527

Per issues reported in PR


kabachuha:
Thanks for the fix! It used to break the API very badly

hithereai:
Cheers!

PR #587, opened 2023-03-13
deprecate `/controlnet/*2img` routes in favor of webui alswayson scripts support in API

### LINKED ISSUE #567: [API-UPDATE] Need to update to accommodate recent API changes in AUTOMATIC1111

### LINKED ISSUE #571: [Bug]: I can't get API call to load any Controlnet process

ljleb:
Continuation of #527. Closes #567, closes #571

I just added the code from earlier PR. I'll update existing tests and add more cases. Let me know if you want me to break the commits into smaller digestible parts.


aiton-sd:
https://github.com/Mikubill/sd-webui-controlnet/pull/586/commits/ed5ac8dc4f50e81ffbb95b166282ae9fa66f6964
Please make this change.

Vespinian:
I'm no expert in this but it doesn't seem like your imports are wrong, but more that it's executing code that is not in a function during the import, seems to be trying to do [this](https://github.com/AUTOMATIC1111/stable-diffusion-webui/blob/master/modules/shared.py#L120). I know it's ugly, but to debug, you could try to put your import inside your functions and see if the behavior changes.

ljleb:
Unfortunately this:

```python
class TestTxt2ImgWorkingBase(unittest.TestCase):
    def setup_route(self, setup_args):
        from modules import scripts
        self.url_txt2img = "http://localhost:7860/sdapi/v1/txt2img"
        self.simple_txt2img = {
            "enable_hr": False,
            "denoising_strength": 0,
            "firstphase_width": 0,\
```

Turns every test case into a failure, with the same error as above/in the failing pipeline. I know the webui imports its own modules in its tests, so I assume it is a configuration error.

Vespinian:
Perphas you could just import the var or class you want from the module. Something like this `from modules.scripts import myClassOrVar`

ljleb:
I tried and it does not work similarly. However, correct me of I'm wrong, but I believe this would not work in case we need to hijack/mock functions or class members I think.

ljleb:
So it seems that some modules can be imported while others cannot. For example, `from modules.paths import script_path` will work while `from modules.scripts import Script` will not work.

The configuration is then probably practically the same for our tests and webui tests.

Vespinian:
But `script_path` is a variable and Script a class no? Maybe `from modules.scripts import scripts_txt2img` would work and give you something workable?

ljleb:
Yeah I tested `scripts_txt2img` and it did not work. I wonder what determines which modules can be imported and which modules cannot.

ljleb:
`modules` does not contain many properties in fact:

image

Vespinian:
What if you make test functions that take no args in the external code library which executes the test you want to make and call those so you only have to import your lib.

I'm kinda joking here but worst case you make api routes for testing and simply do requests in the unit tests to run them.

ljleb:
> @Vespinian worst case you make api routes for testing and simply do requests in the unit tests to run them.

I appreciate your suggestions. We could go there... maybe an alternative solution would be to make our own test base.

After testing, it seems that [this line](https://github.com/AUTOMATIC1111/stable-diffusion-webui/blob/dfeee786f903e392dbef1519c7c246b9856ebab3/launch.py#L351) in the webui is the culprit. Adding `if False and "--skip-torch...` fixes the problem. I think the flag was changed to `--skip-version-check`, maybe we could open a PR to fix this in the webui.

Vespinian:
[This](https://github.com/AUTOMATIC1111/stable-diffusion-webui/commit/a8ff0139637f67af848621968f72e8e620c8e575#diff-cc636c1b90d8799419902fca7cadc64a32239b0e3157e3c4a44f320c298adc84) is the PR that added `--skip-torch-cuda-test` would need to do some repo archeology to find what happened, but it's not in the [shared](https://github.com/AUTOMATIC1111/stable-diffusion-webui/blob/master/modules/shared.py) arg list. Maybe it never was intended to be passed outside the launcher.

ljleb:
I found that adding `--skip-torch-cuda-test` to the cli parser + adding the controlnet extension root to `sys.path` fixes the problem. I opened a PR in the webui for argparser, hoping that this is the correct fix.

For now I'll just set the `IGNORE_CMD_ARGS_ERRORS` env var. It seems that the webui allows to bypass parser errors using this.

Mikubill:
Since this PR contains previously reviewed code, it should be fine to put them together. 
Can you give a summary of diff from #527? Hope I won't mess something up.

ljleb:
Sure, here's the new changelogs:

- tests for `controlnet.image_dict_from_unit` (used by `process` to get a valid image/mask dict from a control unit)
- basic tests for `external_code.update_cn_script_in_place`
- tests for `external_code.get_all_units_from` (used by `process` to convert from flat list to control units)
- more tests for txt2img API. This includes `mask` and `input` properties in control unit json objects
- make old routes `/controlnet/*2img` work again + deprecate them using a warning in console
- reload neighbour modules of `controlnet` when users reload gradio interface. This fixes `external_code` changes not being reflected without a hard restart
- remove `is_img2img` from script positional args. This change affects the API as well, now there is only `is_ui` that still has to be passed around
- move global state and related functions from `controlnet` to `global_state.py`

While writing this, I believe I found a way to infer `is_ui`. As it affects the API I think it should be implemented in this PR... I'll mark as ready, feel free to merge if you don't think it is necessary or can be added later.

ljleb:
Nevermind, the fix I found does not work in batch processing. I don't see anything else to add, if I find something I'll create a new PR.

ljleb:
update: `script.is_img2img` has the right value in `process`, but not in `postprocess`. I used the img2img tab tracker to also keep track of the gradio submit button that has been clicked as a workaround. (i.e. the txt2img button vs the img2img one) I presume this is a bug in the webui? For our purposes, this is good enough. If we eventually really need to know whether we are in img2img or not, ideally `script.is_img2img` should have the right value in all members.

Vespinian:
Did a very simple test real quick by adding prints in process and postprocess of cn. I printed `isinstance(p,StableDiffusionProcessing*2Img)` and `p.scripts is scripts.scripts_*2img`. API returns expected values, but the gradio ui always says I am using the scripts_txt2img. It does pass the right StableDiffusionProcessing for the simple requests I am making. Quite bizzare, but then I found [this](https://github.com/AUTOMATIC1111/stable-diffusion-webui/blob/a9fed7c364061ae6efb37f797b6b522cb3cf7aa2/modules/img2img.py#L154), which would explain my observations but unfortunately make my check unreliable for the ui. Amusingly enough, it calls the right ScriptRunner [a couple lines later](https://github.com/AUTOMATIC1111/stable-diffusion-webui/blob/a9fed7c364061ae6efb37f797b6b522cb3cf7aa2/modules/img2img.py#L169).

On your end, does it only occur when you do the processing switch or all the time?

ljleb:
In `postprocess`, the value of `cn_script.is_img2img` is always `False`, never `True`. Not sure exactly why, as in `process` the script instance variable `is_img2img` and `is_txt2img` take the right values. (i.e. respectively `True` and `False` in img2img, and `False` and `True` in txt2img)

Because of this function:
https://github.com/Mikubill/sd-webui-controlnet/blob/fc904ce6c27be668c745363b9e7ba40a0dda3060/scripts/controlnet.py#L137-L143

Which is called here:
https://github.com/Mikubill/sd-webui-controlnet/blob/fc904ce6c27be668c745363b9e7ba40a0dda3060/scripts/controlnet.py#L821

Using `isinstance` is not reliable either.

In batch img2img, `process` and `postprocess` are called repeatedly, so it's also not reliable to save values in `process` and reset the state of the class in `postprocess`.

ljleb:
As the existing API should still work exactly the same as before, I think we can keep the version set to `1`. I left it at `1` for of this reason.

Vespinian:
If you change [L154](https://github.com/AUTOMATIC1111/stable-diffusion-webui/blob/a9fed7c364061ae6efb37f797b6b522cb3cf7aa2/modules/img2img.py#L154) in the img2img.py of auto1111 from `p.scripts = modules.scripts.scripts_txt2img` to `p.scripts = modules.scripts.scripts_img2img` does it fix your problem?

ljleb:
It seems to be working when making the change you propose. I wonder if there is a problem that using the txt2img scripts fixes? Otherwise, if there is no PR open for this ATM in the webui we should consider opening one.

Vespinian:
In external_code.py, unit type that is passed in `to_processing_unit` is `<class 'scripts.external_code.ControlNetUnit'>` which, for some reason, is not a `ControlNetUnit` if you do an `isinstance`. 

That said, @LonicaMewinsky You can probably just comment out the assert at [L181](https://github.com/Mikubill/sd-webui-controlnet/blob/8682c1e49728926bb7cc7753da8917d1ab095fb1/scripts/external_code.py#L181) of external_code.py as a workaround while this gets looked at.

ljleb:
Ah thanks for the help. Maybe just do a dirty class name check instead.

finley0066:
> Sure, here's the new changelogs:
> 
> * tests for `controlnet.image_dict_from_unit` (used by `process` to get a valid image/mask dict from a control unit)
> * basic tests for `external_code.update_cn_script_in_place`
> * tests for `external_code.get_all_units_from` (used by `process` to convert from flat list to control units)
> * more tests for txt2img API. This includes `mask` and `image` properties in control unit json objects
> * make old routes `/controlnet/*2img` work again + deprecate them using a warning in console
> * reload neighbour modules of `controlnet` when users reload gradio interface. This fixes `external_code` changes not being reflected without a hard restart
> * remove `is_img2img` from script positional args. This change affects the API as well, now there is only `is_ui` that still has to be passed around
> * move global state and related functions from `controlnet` to `global_state` module
> 
> While writing this, I believe I found a way to infer `is_ui`. As it affects the API I think it should be implemented in this PR... I'll mark as ready, feel free to merge if you don't think it is necessary or can be added later.
> 
> Edit: I found a fix after all. Working on it.

Too bad, when synchronizing your latest code, I found that the '/controlnet/*2img' interface is not working, and at the same time, the '/sdapi/v1/*2img' interface is also not working.

finley0066:
can you have a test then do the checkin?

Vespinian:
@finley0066 Is your auto1111 webui up to date too? If it is you might be experiencing this https://github.com/AUTOMATIC1111/stable-diffusion-webui/issues/8666 which should be fixed soon.  If it's not then don't expect anything to work with the latest updates to the controlnet extension, since you might be missing the changes that kicked off this whole rework.

finley0066:
> 



> @finley0066 Is your auto1111 webui up to date too? If it is you might be experiencing this [AUTOMATIC1111/stable-diffusion-webui#8666](https://github.com/AUTOMATIC1111/stable-diffusion-webui/issues/8666) which should be fixed soon. If it's not then don't expect anything to work with the latest updates to the controlnet extension, since you might be missing the changes that kicked off this whole rework.

Yes, I always keep the sd webui up to date, but it ran fine the day before yesterday. Thanks anyway.

paulo-coronado:
Before recent updates my ControlNet API **stopped** working. No errors are displayed and the output is a random image (without ControlNet). According to the docs, the official txt2img route is `/sdapi/v1/txt2img`:

> "API Update: The /controlnet/txt2img and /controlnet/img2img routes are deprecated. Consider [migrating](https://github.com/Mikubill/sd-webui-controlnet/wiki/API#migrating-from-controlnet2img-to-sdapiv12img) to the /sdapi/v1/txt2img and /sdapi/v1/img2img routes instead."

<details>
<summary>Here is the A1111 command I am running</summary>

```bash
python webui.py --port 3000 --nowebui --api --xformers --ckpt models\Stable-diffusion\rpg_v4.safetensors
```

</details>

<details>
<summary>Here is the HTTP endpoint</summary>

```
http://127.0.0.1:3000/sdapi/v1/txt2img
```

</details>

<details>
<summary>Here is the HTTP payload</summary>

```JSON
{
  "prompt": "...",
  "negative_prompt": "...",
  "seed": 123456789,
  "sampler_name": "DDIM",
  "batch_size": 1,
  "steps": 20,
  "cfg_scale": 7,
  "width": 512,
  "height": 512,
  "restore_faces": false,
  "enable_hr": false,
  "tiling": false,
  "alwayson_scripts": {
    "controlnet": {
        "args": [
            {
                "module": "segmentation",
                "model": "control_seg-fp16 [b9c1cc12]",
                "weight": 1,
                "resize_mode": "Envelope (Outer Fit)",
                "lowvram": false,
                "guessmode": false,
                "input_image": "..."
            },
            {
            "module": "canny",
            "model": "control_canny-fp16 [e3fe7712]",
            "weight": 0.5,
            "resize_mode": "Envelope (Outer Fit)",
            "lowvram": false,
            "guessmode": false,
            "input_image": "..."
            }
        ]
    }
  }
}
```

</details>

Can anyone tell me why ControlNet is not being applied? Thanks!

paulo-coronado:
@Vespinian No... it doesn't. CN is not being enabled, because the generated image is the same as if no `alwayson_scripts` was provided.

_**Additional issue**: I think LoRA is also not being applied... when launching `webui.py` I get the following message: "Skipping unknown extra network: lora"._

ljleb:
@paulo-coronado The lora loads without controlnet, is that correct?

I'll try investigating this further, thanks for your detailed report.

Vespinian:
What about launching with the webui? Maybe the arg indices for always_on scripts don't get setup when using --nowebui which changing the api `arg_form` and `arg_to` like you were doing previously would fix.

ljleb:
This looks like the same error message as #603. You are probably running on latest commit, but if you are not please let us know if the recent fix resolves the issue. Also, please open an issue if the problem persists after updating, as it helps us keep track of what work needs to be done in a centralized location in the repo.

Vespinian:
@paulo-coronado Thanks for providing debug info. I was thinking more along the line of just removing `--nowebui` from the lauch args but if your `webui-user.bat` doesn't have it then you've just tried it. But @ljleb is right, we can continue investigating this in an open issue if you are on the latest.

PR #590, opened 2023-03-13
Models recommended for each Preprocessors

brunogcar:
added a list of models recomended for each preprocessor, to be displayed on UI when a preprocessor is selected

Credit @missionfloyd for the code


brunogcar:
just a simple addition to try to clarify specially to new users what goes with what, while i try to learn basic python

message appears below preprocessor and model dropdowns when a preprocessor is selected

removing whitespaces werent intentional

PR #601, opened 2023-03-15
Added  a new API call to get the preprocessor list

mishafarms:
In order to not have to hard code the modules in code that uses the API, it seemed easy to request the list of modules that
can be set.

PR #612, opened 2023-03-17
Revert "Added  a new API call to get the preprocessor list"

ljleb:
Reverts Mikubill/sd-webui-controlnet#601


ljleb:
Apologies @mishafarms I introduced an import issue.

PR #616, opened 2023-03-17
Revert "add back preprocessor in api"

ljleb:
Reverts Mikubill/sd-webui-controlnet#613


ljleb:
Due to some reports that resolution stops working with this change, I'm reverting my changes. Apologies for the inconvenience. I think a dev/master separation as has been proposed in #604 would be useful, we could find these bugs early.

PR #620, opened 2023-03-17
fix inner fit and outer fit

Dradows:


Dradows:
Sorry, I think I shouldn't format the code

PR #621, opened 2023-03-17
Fix inner fit and outer fit

Dradows:


Dradows:
origin image
石头
inner fit 384*768
innerFit1
outer fit 384*768
outerFit1
inner fit 768*384
innerFit2
outer fit 768*384
outerFit2

Dradows:
any problems?

PR #623, opened 2023-03-17
Added a method to get the list of preprocessors

mishafarms:
This is the re-submission of my last PR.

Michael

### DIFF
--- a/scripts/controlnet.py
+++ b/scripts/controlnet.py
@@ -117,12 +117,10 @@ def image_dict_from_unit(unit) -> Optional[Dict[str, np.ndarray]]:
 class Script(scripts.Script):
     model_cache = OrderedDict()
 
-
     def __init__(self) -> None:
         super().__init__()
         self.latest_network = None
         self.preprocessor = global_state.cn_preprocessor_modules
-
         self.unloadable = global_state.cn_preprocessor_unloadable
         self.input_image = None
         self.latest_model_hash = ""
--- a/scripts/controlnet.py
+++ b/scripts/controlnet.py
@@ -18,8 +18,8 @@ from scripts.adapter import PlugableAdapter
 from scripts.utils import load_state_dict
 from scripts.hook import ControlParams, UnetHook
 from scripts import external_code, global_state
-importlib.reload(external_code)
 importlib.reload(global_state)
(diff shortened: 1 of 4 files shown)


mishafarms:
Since you have deprecated the /controlnet/txt2img method. I now get a 500 error. Can you please update the example code?
Michael

ljleb:
Thank you for adding back the PR. Apologies again for the inconvenience.

For the web API, did you see the [migration section](https://github.com/Mikubill/sd-webui-controlnet/wiki/API#migrating-from-controlnet2img-to-sdapiv12img)? The `/controlnet/txt2img` route should be working in 56194e110e62b70f079fbaa0eb7718e0931b9917. If it is not, please open an issue describing the problem so I or someone else can reproduce and fix it.

scripts/controlnet.py
```diff
@@ -790,6 +790,10 @@ def postprocess(self, p, processed, *args):
         gc.collect()
         devices.torch_gc()
 
+
+# Do this so the preprocessor list is available for the API
+global_state.cn_preprocessors_names = Script().preprocessor
```
ljleb, line 795: As there are gradio components created in `__init__`, I am not sure we should be creating a script here. I think we should move the list of preprocessors and unloadables directly to `global_state` instead of replacing the dict here with an updated one.

ljleb: Note that doing this will create a bug for people that want to upgrade without restarting the webui completely. This is because the [`importlib.reload`](https://github.com/Mikubill/sd-webui-controlnet/pull/623/files#diff-86d1966226c5d294b3ed140211dcc872a2e27312f5f9b88c454e4717ef96f596L21-L22) calls are out of order. They should be:

```
importlib.reload(global_state)
importlib.reload(external_code)
```

Instead of the other way around as they are currently.

mishafarms: No problem.

I did not see the migration section. Thank you. I will check the /controlnet/txt2img. 
I had not updated automatic 1111, everything works fine. Thank you.

mishafarms: I will update the code. Hang on.

ljleb:
LGTM. I'll let Mikubill review to make sure we didn't miss anything.

Issue #615, opened 2023-03-17
[Bug]: Txt2img tests do not test anything


ljleb:
### Is there an existing issue for this?

- [X] I have searched the existing issues and checked the recent builds/commits of both this extension and the webui

### What happened?

Debugging into the txt2img api tests, I realized that they fail when trying to call `self.parse_remote_call`. For some reason `args` is a list of 1 list of control unit dicts, which the code does not expect.

Despite this, the tests indicate success. We should change the test behavior so that we can detect failures.

### Steps to reproduce the problem

1. Open an IDE
2. Put a break point inside `process`
3. Debug the txt2img api tests

### What should have happened?

The api tests should fail under the current implementation.

### Commit where the problem happens

controlnet: c9da1a01d92bd1550ee7542617aa205fa71a0b64

### What browsers do you use to access the UI ?

Mozilla Firefox

### Command Line Arguments

```Shell
--no-half --disable-opt-split-attention --use-cpu all --api --tests .\extensions\sd-webui-controlnet\tests --api-log --allow-code
```

### Console logs

```Shell
All tests pass in logs but the extension does not succeed in treating the requests. We should fail try to fail these tests before making them pass to make them more robust.
```

PR #678, opened 2023-03-30
Patched unit module being None in API requests, reworked txt2img test case

Vespinian:
If `module` isn't passed in api requests it would default to `None` and throw an error instead of defaulting to `"none"`. Added a check when parsing the args. 

Tests didn't do requests as expected with the latest args. They were outputting errors in stderr.txt yet passing (Probably because the API request itself wasn't failling, controlnet would just crash and the webui would just march on without it completing the request). A future PR could maybe check stderr and stdout for anomalies. I reworked it so it would construct the alwayson_script param correctly and construct the deprecated txt2img param `controlnet_units` as I understand it.


ljleb:
Related: #615

Thanks @Vespinian for spending time on this! This aspect of the API code is in dire need of attention.

Vespinian:
No problem, having these tests is nice way test changes to the alwayson_scripts behavior.  On another note without https://github.com/AUTOMATIC1111/stable-diffusion-webui/pull/9130 I expect some of the tests to fail should I just open this for review now or do we wait for a possible eventual merge of the PR?

ljleb:
> I expect some of the tests to fail should I just open this for review now or do we wait for a possible eventual merge of the PR? Also can you approve the workflow?

All API tests at the moment silently fail because control units are nested into an additional layer of list. If you un-nest the units in the api call the tests should report failure (that's what happened on my side when I tried to work on it a little). Even though the status code is 200 we should consider fixing these tests.

Sure, I'll aprove the workflow when you mark as ready for review. We could fix the tests in another PR, as long as you do a minimal amount of manual testing.

Vespinian:
Add an experimental stderr file parsing. It seems to work on my local setup, it detects whether or not the None fix is in their. I don't know if it works in the automated workflow.

ljleb:
The pipeline did not fail. If the test for logs works, then it seems all tests pass now.

Vespinian:
If I revert the default module value insert in parse_arg we could test for an error throw. Or I could add some Img2Img tests since the arg resizing bug is sensitive to what is next to the controlnet args in the script_args list. Running Img2Img would probably throw the error.

Vespinian:
Added 3 very basic img2img tests. On my setup, the deprecated route passes for some reason but the alwayson_scripts one throws the error seen in #675. If you patch in https://github.com/AUTOMATIC1111/stable-diffusion-webui/pull/9130 in the webui, the test passes.

Vespinian:
Well turns out the deprecated route test weren't using the deprecated routes and controlnet was therefore unused. Test behavior is now as I expect it on my machine.

ljleb:
One test fails now. It's not 3 but that is progress 👍

Vespinian:
Added 2 tests that modify the `control_net_max_models_num` option since I think the 2 alwayson script test work because it's set to 1. Not sure if it actually works though since the ui function should only be called once that the beginning of all the api tests.

Vespinian:
Well as I sorta semi expected it passed when it shouldn't. Commited a new test that passes 0 units. Hopefully this will trigger a unit to max unit mismatch and throw the error we are looking for in the workflow.

Vespinian:
Successful failure! Probably good enough for now. So now I guess we wait for the fix in the webui.

ljleb:
@Vespinian it appears a1111 has merged your args resizing PR! Cool stuff.

ljleb:
So the tests are fixed now, they pass locally. However, some of the tests depend on the latest version of the webui.

lllyasviel:
a1111 with torch 2.0 is a black hole for github ci. you wont want to face it

ljleb:
The pipeline for a1111 passes on their side on master. Are they doing something special? Or are we doing something special?

lllyasviel:
idk. perhaps they pay more money and have more resources

lllyasviel:
perhaps try replace the commit sha with their lastest sha, instead of remove it

lllyasviel:
well th2.0 is a blcak hole for github ci

ljleb:
The point of this PR is to add tests related to the fixed API code, with alwayson script args resizing. I'll extract the other stuff to another PR, in particular the none fix for unit modules.

lllyasviel:
and you may note that https://github.com/vladmandic/automatic has disabled github ci

lllyasviel:
well perhaps we can find a solution to make it work as auto1111's ci.  let me take a look later

lllyasviel:
interesting. another victim. https://github.com/lshqqytiger/stable-diffusion-webui-directml/commits/master look at their ci tests

ljleb:
This case is different. It goes way past torch2.0. torch seems to be cached? Maybe it is related to the arg parser update.

lllyasviel:
but why they have it cached? how we can have it cached?

ljleb:
> how we can have it cached?

Not sure. Is it in the repo configuration?

lllyasviel:
what about

      - run: |
          pip install torch==2.0.0 torchvision==0.15.1 --extra-index-url https://download.pytorch.org/whl/cu118
          curl -Lo

lllyasviel:
if fail then try 

--index-url https://download.pytorch.org/whl/cu118

and remember to remove the sha we fixed

lllyasviel:
hell

lllyasviel:
ok final solution

pip install torch torchvision torchaudio --index-url https://download.pytorch.org/whl/cu118

lllyasviel:
if this fail then we have no more methods
image

lllyasviel:
gosh it works. but some other line failed

ljleb:
oh wow. nice fix! hope we can get it all to work.

ljleb:
another thing I see once is a while is the setting python 3.10 step getting stuck fetching the cache. when it happens we need to stop and restart the workflow manually. idk why. not at all important but still just a little bit annoying.

ljleb:
I'm looking online for the triton > setuptool dependency. it seems to be the problem here. we could try installing setuptools from the debian package: `pip install python3-setuptools`

ljleb:
possible pip install --upgrade fixes setuptools, let's see the trace of both workflows

ljleb:
oops it's `apt install python3-setuptools` not through pip.

ljleb:
The nature of the issue may be different. The logs have:

```
Successfully installed setuptools-67.7.2
```

Whereas the version triton depends on is >= 40.8.0, clearly it should be a valid candidate. are we using the same python env?

ljleb:
okay you had it right before, there's a pip package: https://pypi.org/project/setuptools/ let's drop apt

lllyasviel:
it wont work i just tried

lllyasviel:
it fails even if the install of setuptool succeed

ljleb:
yes, I don't think this is the issue. or we are not installing it at the right place? very confusing

lllyasviel:
the problem is that the install of that pytorch building cannot use system setuptool, it want to use its own setuptool. but it cannot find a correct setuptool from package index

ljleb:
may be related: https://github.com/pytorch/pytorch/issues/95081

lllyasviel:
extra also wont work, i just tried

ljleb:
ah i missed it my bad. according to the issue above it is supposed to fix the problem

lllyasviel:
oh my gosh it works

ljleb:
> interesting. another victim. https://github.com/lshqqytiger/stable-diffusion-webui-directml/commits/master look at their ci tests

This case is actually not different. We do need `--extra-index-url` after all for pip to be able to download. It's later when launch.py tries to start that it fails.

lllyasviel:
not it is because u changed directory

ljleb:
i'll remove stuff until it breaks earlier

ljleb:
a1111 said they fixed the parser and closed my PR for the fix, in theory we don't need the `IGNORE_CMD_ARGS_ERRORS` env var. if that fixes it well I'll reopen my PR I guess?

ljleb:
no way! looks like you found a fix

ljleb:
quick! merge before the pipeline changes its mind

lllyasviel:
so the problem is that a1111's 2.0.0 is wrong and they should have 2.0.1

lllyasviel:
but this may also work for a period of time as torch will soon remove 2.0.1 for 2.0.2 from index just like they remove 2.0.0

lllyasviel:
well, perhaps just wait for 2.0.2 and then find another fix

ljleb:
are you sure it's not just the torchvision version? the webui seems to have the pipeline work using torch==2.0.0

Issue #502, opened 2023-03-05
[Bug]: StyleAdapter incompatible with --medvram and --lowvram arguments


Danamir:
### Is there an existing issue for this?

- [X] I have searched the existing issues and checked the recent builds/commits of both this extension and the webui

### What happened?

The `clip_vision` and/or `t2iadapter_style` seems to be incompatible with `--medvram` and `--lowvram` command line arguments.

### Steps to reproduce the problem

1. Launch Automatic1111 webui with `--medvram` or `--lowvram` in `COMMANDLINE_ARGS`
2. Enable ControlNet with `clip_vision` preprocessor and `t2iadapter_style` model, add an input style image
2. Generate an image
3. The image generated is the same as one without ControlNet enabled
4. Relaunch the webui without the med/lowvram arguments
5. Generate an image with ControlNet StyleAdapter
6. The image is now generated with the correct style

### What should have happened?

The style of the input image should have been applied.

### Commit where the problem happens

webui: a5a4fb70
controlnet: 84a2b22d 

### What browsers do you use to access the UI ?

Mozilla Firefox

### Command Line Arguments

```Shell
--medvram --xformers --no-half-vae
```

### Console logs

```Shell
Launching Web UI with arguments: --xformers --no-half-vae
Loading model: t2iadapter_style [202e85cc]
Loaded state_dict from [E:\Program Files\stable-diffusion-webui\extensions\sd-webui-controlnet\models\t2iadapter_style.pth]
ControlNet model t2iadapter_style [202e85cc] loaded
100%|██████████████████████████████████████████████████████████████████████████████████| 16/16 [00:01<00:00, 11.60it/s]
Total progress: 100%|██████████████████████████████████████████████████████████████████| 16/16 [00:02<00:00,  7.33it/s]
```

Danamir:
It seems that having more than 75 tokens also breaks the StyleAdapter model.

specblades:
> It seems that having more than 75 tokens also breaks the StyleAdapter model.

Acknowledged. Correspondingly ceases to have an effect when using BREAK as well

Kanareika:
> It seems that having more than 75 tokens also breaks the StyleAdapter model.
Including negative as well.

SeBL4RD:
Can confirm the bug wih --medvram....
Can't use clip_vision without a restart of SD without --medvram, and so, can't do better than 768x768, instead of 1400x1400 with medvram

PR #717, opened 2023-04-07
EXPERIMENTAL: Bugfix for style transfer + BREAK keyword (and >75 tokens).

Symbiomatrix:
Addresses #502, #521, #700, probably. 

The context is received in two separate forward calls, pos and neg, so the extra_cond is appended to each of them as it would normally be, with the assumption that each of (pos,neg) will appear intermittently, and considering ddim/plms will reverse the order of appearance, as I've observed.

NOTE: I have been informed that UNIPC sampler has a similar reversal, should consider adding it to the vanilla list.
NOTE2: The patch has not been thoroughly tested, I'm pretty sure it'll break if combined with batch count. Perhaps a better method of BREAK detection is in order, than the appearance of the warning.

EDIT: I think it actually does work with batch count, surprisingly enough. I'm getting somewhat persistent style and region definitions, and the first gen is identical to batch=1. Not batch size, however, that one completely breaks the style.

RegionStyle2

### DIFF
--- a/scripts/hook.py
+++ b/scripts/hook.py
@@ -69,6 +69,8 @@ class UnetHook(nn.Module):
         super().__init__()
         self.lowvram = lowvram
         self.batch_cond_available = True
+        self.cntiter = -1 # SBM Stylebreak cond / uncond counter.
+        self.stylebreak = False
         self.only_mid_control = shared.opts.data.get("control_net_only_mid_control", False)
         
     def hook(self, model):
@@ -115,6 +117,7 @@ class UnetHook(nn.Module):
             return base + x
 
         def forward(self, x, timesteps=None, context=None, **kwargs):
+            outer.cntiter = outer.cntiter + 1 # SBM Stylebreak attempted fix.
             total_control = [0.0] * 13
             total_adapter = [0.0] * 4
             total_extra_cond = torch.zeros([0, context.shape[-1]]).to(devices.get_device_for("controlnet"))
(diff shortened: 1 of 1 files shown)


scripts/hook.py
```diff
@@ -149,6 +155,25 @@ def forward(self, x, timesteps=None, context=None, **kwargs):
...
+                # SBM Stylebreak fix: Context is split to (cond,uncond) iterations.
+                # Since I don't receieve an indication of which iteration is which,
+                # I rely instead on a simple counter and assumption that the order is maintained.
+                
+                # SBM Perhaps a simple reshape(1,*total.shape) would do.
+                total_extra_cond = torch.repeat_interleave(total_extra_cond.unsqueeze(0), context.shape[0], dim=0)
+                indpos = True
+                if outer.is_vanilla_samplers: # Vanillas reverse cond,uncond order.
+                    indpos = not indpos
+                if outer.cntiter % 2 != 0: # Flip cond/uncond in odd iters.
```
ljleb, line 169: I suspect this is not going to work well with multidiffusion. Some extensions double, triple, quadruple the number of passes sometimes per prompt.

Symbiomatrix: You're probably right about that, but I don't understand the flow well enough to be able to pass a more direct indication.
At the least, it should still work if pos/neg are called intermittently for every pass.

ljleb: Apologies for the delay. It seems to be working with multidiffusion after a bit of testing.

PR #729, opened 2023-04-11
add RGB to BGR in api

cxDosx:


ljleb:
This adds support for rgb to bgr in the deprecated api. Note that you should already be able to use it in the webui alwayson script args api.

PR #742, opened 2023-04-13
Add support for ControlNet 1.1

Mikubill:

PR #785, opened 2023-04-16
Fix shuffle preprocessor seed bug & allow modifying seed with subseed

catboxanon:
Closes https://github.com/Mikubill/sd-webui-controlnet/issues/811 (see additional comment below regarding this fix)

As discussed in https://github.com/Mikubill/sd-webui-controlnet/pull/742, the shuffle processor seed should be reproducible but also controllable by the user. The implementation in https://github.com/Mikubill/sd-webui-controlnet/commit/fa994bb2ffed7428f07afd423f215609c5794204 isn't ideal because when using txt2img, this means that only one variation is able to be created for a generated image.

This PR changes this to also account for the subseed (what the web UI refers to as the `Variation seed`). This means that a user could use this variation seed, with the `Variation strength` set to `0` (so it doesn't affect the base generated image), but it will affect the Shuffle preprocessor to use a different permutation. This will also work with batch sizes out of the box.

A use case for this would be using the XYZ plot script to show how the Shuffle cnet strength affects the image for multiple different variation seeds at increasing strengths, but all starting from the same generated image.

Might be something to reconsider the implementation for later once paint-with-words and GLIGEN support are added. https://github.com/Mikubill/sd-webui-controlnet/discussions/519#discussioncomment-5225020

**Edit:** [See comment below](https://github.com/Mikubill/sd-webui-controlnet/pull/785#issuecomment-1510425473). This now also fixes a bug in the previous implementation.

### DIFF
--- a/scripts/controlnet.py
+++ b/scripts/controlnet.py
@@ -733,7 +733,7 @@ class Script(scripts.Script):
                 input_image = images.resize_image(2, input_image, p.width, p.height)
                 input_image = HWC3(np.asarray(input_image))
 
-            np.random.seed(int(p.seed) % 65535)
+            np.random.seed((int(p.seed) + int(p.subseed)) % 65535)
 
             print(f"Loading preprocessor: {unit.module}")
             preprocessor = self.preprocessor[unit.module]
--- a/scripts/controlnet.py
+++ b/scripts/controlnet.py
@@ -733,7 +733,9 @@ class Script(scripts.Script):
                 input_image = images.resize_image(2, input_image, p.width, p.height)
                 input_image = HWC3(np.asarray(input_image))
 
-            np.random.seed((int(p.seed) + max(int(p.subseed),0)) % 65535)
+            tmp_seed = int(p.all_seeds[0] if p.seed == -1 else max(int(p.seed),0))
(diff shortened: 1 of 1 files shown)


marcsyp:
I sill have this problem, even with the latest webui + CN.  Any ideas why?  Even setting a variation seed (with variation set to 0) does not maintain the shuffle preprocessor stability.  Thanks!

huchenlei:
> I sill have this problem, even with the latest webui + CN. Any ideas why? Even setting a variation seed (with variation set to 0) does not maintain the shuffle preprocessor stability. Thanks!

I just tested that the result for shuffle can be perfectly reproduced in lastest A1111 + ControlNet extension. Make sure to turn on debug log `--controlnet-loglevel DEBUG` and look for following log:
```
2024-04-23 19:44:07,826 - ControlNet - INFO - Loading model from cache: control_v11e_sd15_shuffle [526bfdae]
2024-04-23 19:44:07,826 - ControlNet - DEBUG - is_upscale_script=False
2024-04-23 19:44:07,828 - ControlNet - DEBUG - Use numpy seed 222.
2024-04-23 19:44:07,828 - ControlNet - INFO - Using preprocessor: shuffle
```
The numpy seed should be the same between runs to ensure shuffle preprocessor produce the same result.

image

PR #803, opened 2023-04-16
restore default to inner fit

lllyasviel:


scripts/api.py
```diff
@@ -36,7 +36,7 @@ def encode_np_to_base64(image):
     "module": (str, Field(default="none", title='Controlnet Module')),
     "model": (str, Field(default="None", title='Controlnet Model')),
     "weight": (float, Field(default=1.0, title='Controlnet Weight')),
-    "resize_mode": (Union[int, str], Field(default="Outer Fit (Shrink to Fit)", title='Controlnet Resize Mode')),
+    "resize_mode": (Union[int, str], Field(default="Inner Fit (Scale to Fit)", title='Controlnet Resize Mode')),
```
ljleb, line 39: Note that you can use an int instead now

PR #807, opened 2023-04-16
Cache preprocessor results

asagi4:
I noticed that running some ControlNet preprocessors can take  quite a long time when using scripts (like X/Y/Z grid)
This adds a simple cache to avoid needlessly running the annotator every time if the input image doesn't change.

It shaves off several seconds of delay between iterations, at least on my computer.

I don't know if all preprocessors are cacheable. Shuffle isn't, so I excluded it. It's fast enough anyway.

Dunno if the default size should be nonzero. I set it that way to preserve current behaviour.

### DIFF
--- a/scripts/controlnet.py
+++ b/scripts/controlnet.py
@@ -206,6 +206,7 @@ class UiControlNetUnit(external_code.ControlNetUnit):
 
 class Script(scripts.Script):
     model_cache = OrderedDict()
+    preprocessor_cache = {}
 
     def __init__(self) -> None:
         super().__init__()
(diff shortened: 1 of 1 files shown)


lllyasviel:
how this works? can u expain a bit?

asagi4:
@lllyasviel It memoizes the preprocessor function with an in-memory cache so that if an annotator would be called with identical parameters (ie. same input image with the same parameters) twice, the second call will take the annotator result from a memory cache instead of actually running the annotator.

It doesn't matter for all preprocessors since they're quick enough anyway, but some of them can take several seconds to run.

The cache key is created by turning the input image and parameters into a string that it uses as part of the cache key. It's hacky, but as far as I could tell, this works just fine as long as the annotators are dependent on only the input parameters (shuffle isn't, hence it's excluded)

lllyasviel:
it sounds like official python's implemetation has same thing
https://docs.python.org/3/library/functools.html
can we use this?

asagi4:
I don't think the decorator will work because using an image directly as a dictionary key doesn't seem to be possible; that's why I'm converting it into a string. Creating the wrapper manually also allows users to adjust the maximum cache size at runtime.

lllyasviel:
it seems that this will introduce some complicated logic and needs more tests

asagi4:
What do you think is complicated about it? It just wraps the preprocessor function, and if the cache size is set to 0 (the default), it's a no-op.

There's theoretically a chance that some `(model, image, args)` tuples might result in the same cache key, but I'm fairly sure that's very unlikely to happen in practice.

Granted, I don't know if all preprocessors are cacheable, but as far as I can tell, only shuffle did anything random or seed-dependent.

lllyasviel:
it seems that this can be beneficial to xyz girds but will add more processing to all standard use cases.
considering this can influence all use cases, we will need more tests about it (and in the future our debugging will have another layer in the calling stack)

asagi4:
How would you like it to be tested? I've been using it myself for a while, and for the "normal" use-cases the overhead of the cache hasn't been noticeable. The amount of data in the cache is in the tens of megabytes at most depending on how large you make it and how large the images in it are, and the wrapper itself has virtually no performance impact since preprocessor calls are not on a hot path.

lllyasviel:
we will see if this is safe enough and will not bring more things like https://github.com/Mikubill/sd-webui-controlnet/issues/907 and https://github.com/Mikubill/sd-webui-controlnet/discussions/884,
adding steps before preprocessors is considered as dangerous after we devoted lots of efforts to handle those problems

huchenlei:
> I don't think the decorator will work because using an image directly as a dictionary key doesn't seem to be possible; that's why I'm converting it into a string. Creating the wrapper manually also allows users to adjust the maximum cache size at runtime.

In `functools.lru_cache` [implementation](https://github.com/python/cpython/blob/2866e030f01dc3ff08de32857fa77d52468b676b/Lib/functools.py#L477), it is calling `hash` on input parameters.

You can just use LRU cache decorator instead of reinventing the wheel.

lllyasviel:
wait a moment, is `str(image)` just equal to something like `str(image)=="np.ndarray([[[XXXX....XXXX]]])"`
then it is clearly not correct technically

huchenlei:
> wait a moment, is `str(image)` just equal to something like `str(image)=="np.ndarray([[[XXXX....XXXX]]])"` then it is clearly not correct technically if we have an OJ then this is a WA

Yes, `str(image)` will omit information. Calling `image.tostring()` is probably the right way to do it.

asagi4:
@lllyasviel I don't understand your last comment? What do you mean with OJ and WA? I'm confused.

Should I fix the hashing and resubmit, or what do you want done?

PR #841, opened 2023-04-17
flatten ui args with sentinel

### LINKED ISSUE #812: [Bug]: Weight, Guidance Start and Guidance End Selected isn't always being correctly used

### What happened?

When generating an image after changing the Weight, Guidance Start and Guidance End, the Weight, Guidance Start and Guidance End valuez shown in the prompt information is incorrect.

For example you can select 1 and generate and the prompt information will show 1, then you change the value to 0.9 and generate, but the prompt information will now show 0.95.

The same bug happens with all 3 value types.

It seems inconsistent however.

1. Use any ControlNet model to generate an image
image
` ControlNet-0 Enabled: True, ControlNet-0 Module: none, ControlNet-0 Model: control_v11p_sd15_lineart [43d4be0d], ControlNet-0 Weight: 1, ControlNet-0 Guidance Start: 0, ControlNet-0 Guidance End: 1`
2. Change one of the weight, guidance start or guidance end values.
image
3. Check the prompt information and the value does not match what is selected (It may require a few attempts to get it to do this)
`ControlNet-0 Enabled: True, ControlNet-0 Module: none, ControlNet-0 Model: control_v11p_sd15_lineart [43d4be0d], ControlNet-0 Weight: 0.9, ControlNet-0 Guidance Start: 0, ControlNet-0 Guidance End: 1`

The value in the prompt information should reflect the selected value, from testing it also seems that the generation is using the incorrect value and not the one chosen in the ControlNet UI.

webui: 22bcc7be428c94e9408f589966c2040187245d81
controlnet: 82c42285b9e28f6ab76aa35e2ed62690639b2b5c

_No response_

### Command Line Arguments

```Shell
--opt-sdp-attention --no-half-vae --opt-channelslast
```

### Console logs

```Shell
Nothing different happens in the console logs between the generations
```

### Additional information

_No response_

### LINKED ISSUE #821: [Bug]: Image isn't cleared properly causing bad generations after the settings/images/models are changed.

### What happened?

Might be related to: This might be related to https://github.com/Mikubill/sd-webui-controlnet/pull/642   
  
I have noticed bad generations in extended sessions using CN (even pre-1.1). I think controlnet occasionally fails to properly clear an image which results in bad generations from that point forward in the session.  
   
An obvious way to see something isn't working as it should, is to start a new session w/ A1111, enter a simple prompt, put "canny" for a pre-processor, "canny" for the model, but don't put an image, click generate. In the console you'll see "ValueError: controlnet is enabled but no input image is given". Control net doesn't influence the generated image (as intended) - you get a pure prompt based image w/o CN applied. 

Now, add an image, leave preprocessor canny/Model canny and click generate. Generate runs as it should and CN makes the proper changes to the output. 

X out the image so it's removed and there isn't one in the module. Click generate again w/ Cany/Cany still set and again no image as before. This time there is no error in the console and there is an image generated that CN has modified. From this point forward the generations are incorrect regardless of images added / changes to preprocessor or model.

I haven't tested enough scenarios to see if there are other ways for this seeming issue to have something left that messes up future generations - but I know it's annoyed me for a while - I just never noticed a way to see something potentially causing it - the console not responding as it should to the no image the second time around, and then generations changing seems to pint to the problem. 

1. Launch A1111 (Current commit)
2. Enter a simple prompt ("Paris" for example) / Leave everything else default
3. Enable Control net and set Canny for the Preprocessor and Model but do not add an image. 
4. Click generate and notice the error in the console regarding No image and that the image generated isn't altered by CN at all.
5. Add an image and leave everything else the same
6. Click generate and notice an image is generated w/ CN manipulating it as expected.
7. Click [X] to remove the image / Should be back to canny/canny set but no image again and everything else the same.
8. Click generate and notice there is no error in the console and an image is generated that has been altered by CN.
9. Add a new image and click generate - The generated image is not correct (wouldn't match what you'd get if you hadn't clicked generate w/o an image the 2nd time)
10. Repeat the process but don't do step 4 (try to generate when an image isn't in the CN console) 
11. Compare generations from that point forward between the two sessions and notice that the lack of error/awareness that an image wasn't there the first time around causes incorrect generations from then on. 

When there is no image in the console there should always be an error when generate pressed. Prior images shouldn't be held when there isn't an image visible in the console. Proper image generation throughout a session. 

webui:  : 22bcc7be
controlnet: a946afb3 

Google Chrome

### Command Line Arguments

```Shell
-opt-sdp-no-mem-attention --no-half-vae --opt-channelslast 

(But I've tried w/ no arguments and also w/ XFormers)
```

### Console logs

```Shell
venv "E:\stable-diffusion-webui\venv\Scripts\Python.exe"
Python 3.10.6 | packaged by conda-forge | (main, Oct 24 2022, 16:02:16) [MSC v.1916 64 bit (AMD64)]
Commit hash: 22bcc7be428c94e9408f589966c2040187245d81



Launching Web UI with arguments: --opt-sdp-no-mem-attention --no-half-vae --opt-channelslast --skip-version-check --ckpt-dir e:\Stable Diffusion Checkpoints
CUDNN Benchmark Disabled
Enabled CUDA & CUDNN TF32 Sucessfully
CUDA Matmul fp16/bf16 Reduced Precision Reduction Disabled
No module 'xformers'. Proceeding without it.
E:\stable-diffusion-webui\venv\lib\site-packages\torchvision\transforms\functional_tensor.py:5: UserWarning: The torchvision.transforms.functional_tensor module is deprecated in 0.15 and will be **removed in 0.17**. Please don't rely on it. You probably just need to use APIs in torchvision.transforms.functional or in torchvision.transforms.v2.functional.
  warnings.warn(
Loading weights [cc6cb27103] from e:\Stable Diffusion Checkpoints\v1-5-pruned-emaonly.ckpt
Creating model from config: E:\stable-diffusion-webui\configs\v1-inference.yaml
LatentDiffusion: Running in eps-prediction mode
DiffusionWrapper has 859.52 M params.
Loading VAE weights specified in settings: E:\stable-diffusion-webui\models\VAE\vae-ft-mse-840000-ema-pruned.ckpt
Applying scaled dot product cross attention optimization (without memory efficient attention).
Textual inversion embeddings loaded(0):
Model loaded in 3.2s (load weights from disk: 1.0s, create model: 0.2s, apply weights to model: 0.3s, apply channels_last: 0.2s, apply half(): 0.3s, load VAE: 0.2s, move model to device: 0.4s, load textual inversion embeddings: 0.6s).
Running on local URL:  http://127.0.0.1:7860

To create a public link, set `share=True` in `launch()`.
Startup time: 6.8s (import torch: 0.8s, import gradio: 0.5s, import ldm: 0.3s, other imports: 0.5s, load scripts: 0.7s, load SD checkpoint: 3.2s, create ui: 0.7s).
Loading model: control_v11p_sd15_canny [d14c016b]
Loaded state_dict from [E:\stable-diffusion-webui\extensions\sd-webui-controlnet\models\control_v11p_sd15_canny.pth]
ControlNet model control_v11p_sd15_canny [d14c016b] loaded.
Error running process: E:\stable-diffusion-webui\extensions\sd-webui-controlnet\scripts\controlnet.py
  File "E:\stable-diffusion-webui\modules\scripts.py", line 417, in process
  File "E:\stable-diffusion-webui\extensions\sd-webui-controlnet\scripts\controlnet.py", line 731, in process
    raise ValueError('controlnet is enabled but no input image is given')
ValueError: controlnet is enabled but no input image is given

100%|██████████████████████████████████████████████████████████████████████████████████| 20/20 [00:01<00:00, 14.08it/s]
Total progress: 100%|██████████████████████████████████████████████████████████████████| 20/20 [00:00<00:00, 25.65it/s]
Loading model from cache: control_v11p_sd15_canny [d14c016b]███████████████████████████| 20/20 [00:00<00:00, 36.24it/s]
Loading preprocessor: canny
100%|██████████████████████████████████████████████████████████████████████████████████| 20/20 [00:00<00:00, 20.84it/s]
Total progress: 100%|██████████████████████████████████████████████████████████████████| 20/20 [00:01<00:00, 18.19it/s]
Loading model from cache: control_v11p_sd15_canny [d14c016b]███████████████████████████| 20/20 [00:01<00:00, 25.11it/s]
Loading preprocessor: canny
100%|██████████████████████████████████████████████████████████████████████████████████| 20/20 [00:01<00:00, 19.01it/s]
Total progress: 100%|██████████████████████████████████████████████████████████████████| 20/20 [00:01<00:00, 17.66it/s]
Loading model from cache: control_v11p_sd15_canny [d14c016b]███████████████████████████| 20/20 [00:01<00:00, 24.38it/s]
Loading preprocessor: canny
100%|██████████████████████████████████████████████████████████████████████████████████| 20/20 [00:00<00:00, 20.61it/s]
Total progress: 100%|██████████████████████████████████████████████████████████████████| 20/20 [00:01<00:00, 18.29it/s]
Loading model from cache: control_v11p_sd15_canny [d14c016b]███████████████████████████| 20/20 [00:01<00:00, 25.14it/s]
Loading preprocessor: canny
100%|██████████████████████████████████████████████████████████████████████████████████| 20/20 [00:00<00:00, 24.58it/s]
Total progress: 100%|██████████████████████████████████████████████████████████████████| 20/20 [00:01<00:00, 18.34it/s]
Loading model from cache: control_v11p_sd15_canny [d14c016b]███████████████████████████| 20/20 [00:01<00:00, 25.38it/s]
Loading preprocessor: canny
100%|██████████████████████████████████████████████████████████████████████████████████| 20/20 [00:01<00:00, 19.97it/s]
Total progress: 100%|██████████████████████████████████████████████████████████████████| 20/20 [00:01<00:00, 18.28it/s]
Loading model from cache: control_v11p_sd15_canny [d14c016b]███████████████████████████| 20/20 [00:01<00:00, 25.13it/s]
Loading preprocessor: canny
100%|██████████████████████████████████████████████████████████████████████████████████| 20/20 [00:01<00:00, 18.82it/s]
Total progress: 100%|██████████████████████████████████████████████████████████████████| 20/20 [00:01<00:00, 17.08it/s]
Loading model from cache: control_v11p_sd15_canny [d14c016b]███████████████████████████| 20/20 [00:01<00:00, 24.47it/s]
Loading preprocessor: canny
100%|██████████████████████████████████████████████████████████████████████████████████| 20/20 [00:00<00:00, 24.38it/s]
Total progress: 100%|██████████████████████████████████████████████████████████████████| 20/20 [00:01<00:00, 18.01it/s]
Total progress: 100%|██████████████████████████████████████████████████████████████████| 20/20 [00:01<00:00, 25.01it/s]
```

### Additional information

I had issues like this for a while so I stuck w/ an early commit of the pre 1.1 CN. With the release of the new 1.1 models I updated and I have this problem with bad generations that I finally think I found a cause for.

ljleb:
closes #821
closes #812


Mikubill (approved):

ljleb:
I'll merge but consider reverting if it creates unpredicted issues.

lllyasviel:
i think we have problems. now i always need to click button twice to update cn parameters.

lllyasviel:
perhaps i know what is going on. when multiple functions are binded to same gradio buttons, the order of excution is not promised

lllyasviel:
how we can fix this?

ljleb:
I don't know really. Maybe reverting this could be a good start. The problem is that we cannot pass 15 * max_num_of_cnet because the web API will become unusable (users will have to fill in a stupid number of `None`s or something).

So we need to be able to pass gr.States of control units in one way or another, either dicts or ControlNetUnit. Then, whenever a knob is tweaked, we can update the gr.State. The problem with this approach is that some of the values (notably images) are not appropriately updated before generation starts. Also if we update the controls too fast, the values are wrong.

ljleb:
Gradio is just not cut out for this type of project, with the usage restrictions imposed by the webui programming interface. We should look for the least bad solution instead of trying to fix it completely I think... It does not seem like this is 100% resolvable.

PR #849, opened 2023-04-17
better hint

lllyasviel:


Mikubill:
Hint is a bit too long - It might be better to make it shorter for readability

Issue #809, opened 2023-04-16
[Help Needed] Someone to maintain the preprocessor API


lllyasviel:
I am not very familiar with the API part and I am mainly maintaining the correctness of inference.
We need some other contributors to maintain the correctness of APIs for other projects to use.

lllyasviel:
@Mikubill can you take a look at api part?

physis123:
I can't really help I'm afraid but I am hoping the new depth_leres parameters (Remove Near % and Remove Background %) will be added to ControlNetUnitRequest in the API soon.

ljleb:
We should `getattr` on `scripts.processor` and only list exceptional cases with `if elif` cases, otherwise this will be time consuming to maintain.

ljleb:
Should be fixed now.

PR #850, opened 2023-04-17
forward web api to external_code api

ljleb:
Centralizes the preprocessors and unloadables to one place.

@lllyasviel you may be interested in reviewing this.

Related: #809


kft334:
Line 219 should probably be changed to:
results.append(processor_module(img=img, res=controlnet_processor_res, thr_a=controlnet_threshold_a, thr_b=controlnet_threshold_b)[0])

I'm getting empty success responses without the index.

lllyasviel:
it seems conflict the new added mediapipe face. can we fix?

ljleb:
merged main into this. I am curious whether the preprocessor names will be appropriate? I am not sure why we had this huge list of `if`-`elif`s. In any case I think we do not want to have to use different model names through the web api vs the gradio interface or the external code api.

ljleb:
Alias names are not taken into consideration right now, working on a fix

PR #855, opened 2023-04-17
just a quick fix for current f16 rename problem

revolverocelot1:
this error "AssertionError: Error: The model config /content/stable-diffusion-webui/extensions/sd-webui-controlnet/models/control_v11p_sd15_canny_fp16.yaml is missing. ControlNet 1.1 must have configs.", just renames it and model works as expected in local and collab.


revolverocelot1:
> i think both of the ymal files should remain, a bit messy but it'll work

revolverocelot1:
> This works well, it fixes the problem. I am however worried about having to change multiple configs if we need to update any of them. Also, it may encourage followup PRs to do the same for missing yaml files, which would make the situation a little bit less manageable every time.
> 
> Maybe I am overthinking but this is why I think we need a more general solution.
> 
> In the end I'll let Mikubill decide, I think they have a better idea of whether we need to resort to a more general solution than duplicating the configs.

understood, this just a supposed to be small fix not replacement. I still have no idea what is causing the problem to make general solution

lllyasviel:
fixed https://github.com/Mikubill/sd-webui-controlnet/commit/3bbdf73f0024ed93268a65a4dae98416762c98c1

PR #863, opened 2023-04-17
fix: Raise 500 exception instead of returning an error on `detect` API call

ramyma:
I noticed that the `detect` API call returns with a `200` status code when the selected module is not implemented/available or no when no input image is provided.
This PR raises an 500 status code HTTP exception for better integration with 3rd party implementations.

Note: I couldn't find a PEP formatting guide on the repo, so I didn't format the file on saving to avoid having unrelated changes on the PR.

### DIFF
--- a/scripts/api.py
+++ b/scripts/api.py
@@ -209,53 +209,7 @@ def controlnet_api(_: gr.Blocks, app: FastAPI):
             raise HTTPException(
                 status_code=500, detail="Module not available")
         
-        available_modules = [
-            "none",
-
-            "canny",
-
-            "depth_midas",                  # Unload
-            "depth_leres",                  # Unload
-            "depth_zoe",                    # Unload
-
-            "lineart",                      # Unload
-            "lineart_coarse",               # Unload
-            "lineart_anime",                # Unload
-
-            "mediapipe_face",
-
-            "mlsd",                         # Unload
-
-            "normal_midas",                 # Unload
-            "normal_bae",                   # Unload
-
-            "openpose",                     # Unload
-            "openpose_face",                # Unload
-            "openpose_faceonly",            # Unload
-            "openpose_hand",                # Unload
-            "openpose_full",                # Unload
-
-            "scribble_hed",
-            "scribble_pidinet",
-            "scribble_xdog",
-
-            "seg_ofcoco",                   # Unload
-            "seg_ofade20k",                 # Unload
-            "seg_ufade20k",                 # Unload
-
-            "shuffle",
-
-            "softedge_hed",                 # Unload
-            "softedge_hedsafe",             # Unload
-            "softedge_pidinet",             # Unload
-            "softedge_pidisafe",            # Unload
-
-            "t2ia_color_grid",
-            "t2ia_sketch_pidi",
-
-            "threshold"
-        ]
-
+       
         if len(controlnet_input_images) == 0:
             raise HTTPException(
                 status_code=500, detail="No image selected")
--- a/scripts/api.py
+++ b/scripts/api.py
@@ -2,6 +2,7 @@ from typing import Union
 
 import numpy as np
 from fastapi import FastAPI, Body
+from fastapi.exceptions import HTTPException
 from PIL import Image
 import copy
 import pydantic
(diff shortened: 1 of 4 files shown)


huchenlei:
Can you add an unittest under `tests/web_api` for this change? Thanks!

ramyma:
Added tests and rebased on main branch.

scripts/api.py
```diff
@@ -205,9 +206,59 @@ async def detect(
         controlnet_module = global_state.reverse_preprocessor_aliases.get(controlnet_module, controlnet_module)
 
         if controlnet_module not in global_state.cn_preprocessor_modules:
-            return {"images": [], "info": "Module not available"}
+            raise HTTPException(
+                status_code=500, detail="Module not available")
+        
+        available_modules = [
```
ljleb, line 212: is there a reason we need to duplicate the preprocessors keys here? The exception above should handle all cases.

ljleb: Ah it seems to be an artifact from older code. In any case, please consider removing it.

ramyma: You're right, it was pulled in with the rebase for some reason!
I'll remove it.

ramyma:
I rebased again with the latest and fixed a couple of mistakes in the tests.

I wish there were an easier way to run the tests as it takes quite a long time to spin up the whole webui environment through `launch.py`.

tests/web_api/detect_test.py
```diff
@@ -0,0 +1,39 @@
...
+        }
+
+    def test_detect_with_invalid_module_performed(self):
+        detect_args = self.base_detect_args.update({
+            "controlnet_module": "INVALID",
+        })
+        self.assertEqual(utils.detect(detect_args).status_code, 500)
+
+    def test_detect_with_no_input_images_performed(self):
+        detect_args = self.base_detect_args.update({
```
ljleb, line 26: update returns `None` and updates the dict in place. Maybe just use `self.base_detect_args` directly in the assert below instead.

ramyma: Looks like my python is a bit rusty :)
I've switched it up to handle it without mutating the base args.

tests/web_api/detect_test.py
```diff
@@ -0,0 +1,39 @@
...
+        self.base_detect_args = {
+            "controlnet_module": "canny",
+            "controlnet_input_images": [utils.readImage("test/test_files/img2img_basic.png")],
+            "controlnet_processor_res": 512,
+            "controlnet_threshold_a": 0,
+            "controlnet_threshold_b": 0,
+        }
+
+    def test_detect_with_invalid_module_performed(self):
+        detect_args = self.base_detect_args | {
```
ljleb, line 20: Although this is a good solution for python 3.10, this makes the test not work under python 3.8 as the dict union operator was introduced in python 3.9. It was discussed earlier that the repo should support both 3.8 and 3.10.

As this is code that will only be run during automated testing, this may be okay. However, this means we now can only run the tests using 3.10 now.

ramyma: I wasn't aware the repo should support 3.8 as well.
I pushed another update to cover it.
It would be nice if there's a contribution guide, maybe it's there already and I missed.

tests/web_api/detect_test.py
```diff
@@ -0,0 +1,39 @@
...
+        self.base_detect_args = {
+            "controlnet_module": "canny",
+            "controlnet_input_images": [utils.readImage("test/test_files/img2img_basic.png")],
+            "controlnet_processor_res": 512,
+            "controlnet_threshold_a": 0,
+            "controlnet_threshold_b": 0,
+        }
+
+    def test_detect_with_invalid_module_performed(self):
+        detect_args = self.base_detect_args.copy().update({
```
ljleb, line 20:
```suggestion
        detect_args = self.base_detect_args.copy()
        detect_args.update({
```

tests/web_api/detect_test.py
```diff
@@ -0,0 +1,39 @@
...
+            "controlnet_module": "INVALID",
+        })
+        self.assertEqual(utils.detect(detect_args).status_code, 500)
+
+    def test_detect_with_no_input_images_performed(self):
+        detect_args = self.base_detect_args.copy().update({
```
ljleb, line 26:
```suggestion
        detect_args = self.base_detect_args.copy()
        detect_args.update({
```

tests/web_api/detect_test.py
ljleb: No problem, thanks for your time updating the code.

ramyma:
@ljleb  I was just about to push the same changes that you pushed.
It's kinda late and I missed it :)
Thanks for the support!

ljleb:
Ah apologies, I didn't see the changes were recent. Should have waited.

Would it be possible to use http code 422 instead of 500? 500 is usually more employed for internal server errors, as opposed to bad request parameters or body.

ljleb:
It may be a good time to add a contributing.md file so that contributors know what to do and we don't have to tell everybody these things all the time...

ramyma:
Done!

ljleb:
Thanks for your contribution! I'll test the code now.

PR #866, opened 2023-04-17
major update - ControlNet 1.1 GUI

lllyasviel:


ljleb:
Hmm I think we need RGB to BGR when using a manual map. Would be nice to have a preprocessor for it.

Issue #878, opened 2023-04-18
[Bug]: Except for the Canny model, all other models are giving errors.


DivineYun:
### Is there an existing issue for this?

- [X] I have searched the existing issues and checked the recent builds/commits of both this extension and the webui

### What happened?

In v1.1，Except for the canny model, all other models cannot be used and will always prompt that there is a problem with the model.

I have tried deleting the folder and re-downloading, as well as upgrading directly from an old version to a new version. Of course, I have also tried restarting the UI and even restarting the computer, but none of these have solved the problem.

I have posted the code here. Please take a look, experts.

### Steps to reproduce the problem

There are no special steps, the error occurs during normal usage. I have been using ControlNet for a long time before this, so there is no issue of not knowing how to use it.

1. Click "Enable" and "No Prompt Word Mode" (I only did this for testing purposes, even if there are prompt words, the error will still occur.)
2. Import the image.
3. Select the processor and model.
4. Adjust the image size.
5. Preview the processing result (an error will occur, please see the first paragraph of the error code below).
6. Click "Generate Image" (an error will occur, please see the second paragraph of the error code below).

### What should have happened?

none

### Commit where the problem happens

I have used two computers and tested with different versions of webui, but the same error occurred on both.
I use win11.

1、webui: python: 3.10.10  •  torch: 1.13.1+cu117  •  xformers: 0.0.16rc425  •  gradio: 3.23.0 

2、webui: python: 3.10.11  •  torch: 2.0.0+cu118  •  xformers: N/A  •  gradio: 3.23.0  
controlnet: v1.1

### What browsers do you use to access the UI ?

Microsoft Edge

### Command Line Arguments

```Shell
Launching Web UI with arguments: --autolaunch --xformers --no-half-vae --disable-nan-check
```

### Console logs

```Shell
Python 3.10.11 (tags/v3.10.11:7d4cc5a, Apr  5 2023, 00:38:17) [MSC v.1929 64 bit (AMD64)]
Commit hash: 22bcc7be428c94e9408f589966c2040187245d81
Installing requirements for Web UI

Installing sd-dynamic-prompts requirements.txt

Installing sd-webui-controlnet requirement: mediapipe==0.9.1.0

Installing sd-webui-controlnet requirement: opencv-python

Launching Web UI with arguments: --autolaunch --api --disable-nan-check
No module 'xformers'. Proceeding without it.

Civitai Helper: Get Custom Model Folder
[...]

PR #881, opened 2023-04-18
change annotator paths so that users can remove all when corrupted

lllyasviel:
more and more users complain about that they cannot use some preprocessors.
the reason is that the models are corrupted during downloads.
very disappointing that even if they completely remove controlnet, corrupted models cannot be removed, because they are in webui/models/XXX
all annotators should be in our folder so that in the worst case, users can always delete controlnet and install again to handle all problems.
I am testing it now. will merge when test finish


ljleb:
I am not sure about this. What if I want to put these models on a ssd or a different drive instead? They can take up a lot of space together.

lllyasviel:
but how we can help people like https://github.com/Mikubill/sd-webui-controlnet/issues/878
in some country the internet is not very good and model corruption is common
they will be disappointed when they cannot solve problems by removing controlnet and install it again

ljleb:
Yes I agree. I think it's okay to change the defaults.

With the current changes here, everyone will have to download the preprocessor models twice. I was thinking there should be a less radical way to do this transition.

lllyasviel:
"deleting sd-webui-controlnet folder and install again can solve all problems" should be the bottom line.
costumization and backwards compatibility are important, but not deadly
this is deadly

lllyasviel:
a  “less radical way” does not exist because their files maybe already corrpted
detecting each file is not practical and not beneficial for future extending from community because it make things more complicated

sanshaoyeyang:
We are stuck with this problem, please help

sanshaoyeyang:
  File "J:\AI\stable-diffusion-webui\stable-diffusion-webui\venv\lib\site-packages\gradio\routes.py", line 394, in run_predict
    output = await app.get_blocks().process_api(
  File "J:\AI\stable-diffusion-webui\stable-diffusion-webui\venv\lib\site-packages\gradio\blocks.py", line 1075, in process_api
    result = await self.call_function(
  File "J:\AI\stable-diffusion-webui\stable-diffusion-webui\venv\lib\site-packages\gradio\blocks.py", line 884, in call_function
    prediction = await anyio.to_thread.run_sync(
  File "J:\AI\stable-diffusion-webui\stable-diffusion-webui\venv\lib\site-packages\anyio\to_thread.py", line 31, in run_sync
    return await get_asynclib().run_sync_in_worker_thread(
  File "J:\AI\stable-diffusion-webui\stable-diffusion-webui\venv\lib\site-packages\anyio\_backends\_asyncio.py", line 937, in run_sync_in_worker_thread
    return await future
  File "J:\AI\stable-diffusion-webui\stable-diffusion-webui\venv\lib\site-packages\anyio\_backends\_asyncio.py", line 867, in run
    result = context.run(func, *args)
  File "J:\AI\stable-diffusion-webui\stable-diffusion-webui\extensions\sd-webui-controlnet\scripts\controlnet.py", line 390, in run_annotator
    result, is_image = preprocessor(img, res=pres, thr_a=pthr_a, thr_b=pthr_b)
  File "J:\AI\stable-diffusion-webui\stable-diffusion-webui\extensions\sd-webui-controlnet\scripts\processor.py", line 205, in openpose_face
    result = model_openpose(img, include_hand=False, include_face=True)
  File "J:\AI\stable-diffusion-webui\stable-diffusion-webui\extensions\sd-webui-controlnet\annotator\openpose\__init__.py", line 79, in __call__
    self.face_estimation.model.to(self.device)
AttributeError: 'NoneType' object has no attribute 'model'

ljleb:
Can't we just add a checkbox in the settings to change the annotator path? check it by default for new installs, and do not check it for existing installs.

lllyasviel:
wait a minute, where clip_vision is downloaded?

lllyasviel:
clip vision added. testing now

lllyasviel:
yes we need a config but the starting point should be a case that "deleting sd-webui-controlnet folder and install again can solve all problems" 
we must achieve this in the first place, and this is emergency.
after this we can start thinking about other improvements

ljleb:
I think this needs a bit more tinkering, it seems that the old model path is loaded by default if it exists:

```
https://github.com/Mikubill/sd-webui-controlnet/blob/0696c67cfef2bb991157470475583c3c7c8b45ae/annotator/leres/__init__.py#L42
```

PR #886, opened 2023-04-18
use text consistent to a1111 resizing

lllyasviel:


ljleb:
The change in the texts for `ResizeMode` will break existing api code. Users often use the strings, we should add aliases to old names.

ljleb:
I'll finish up the paths settings and then I'll add the aliases.

PR #894, opened 2023-04-18
resize mode aliases

ljleb:


omair-kg:
The quickest fix is to change the keys of the aliases dict

aarlo:
I'm getting that error too! Where is that aliases dict?

omair-kg:
scripts/external_code.py
image

PR #903, opened 2023-04-19
super high quality resampling

lllyasviel:
This a super high-quality sampling for image resizing and high-res resizing.
it keep segmentation maps as segmentation maps with arbitary resizing 
it keep canny maps as canny maps with arbitary resizing 
it keep mlsd maps as mlsd maps with arbitary resizing 
it keep scribble maps as scribble maps with arbitary resizing 

also, now if users use high-res fix, the both two control maps in low res and high res will be visualized so that users know the two passes


lllyasviel:
it work like this

this is low res canny for 1st low resolution diffusion
image

this is canny after the "super high quality resampling" for high-res fix
image

note that the second image is compued from first image.
it is not soemthing like detect canny maps again (we should not detect canny maps since users cannot control this and some images may have low res and detect again is bad, also users can use none as preprocessor)

this also improves seg mlsd and scribbles

lllyasviel:
now users can use very high denoising rate in even high-res pass

lllyasviel:
I am still testing this. will merge after test

lllyasviel:
seg, low pass
image

seg, high pass
image

lllyasviel:
mlsd low pass
image

mlsd high res fix
image

lllyasviel:
it seems that mlsd part need revision, doing it now

lllyasviel:
fixed now

mlsd first pass
image

mlsd high res fix
image

lllyasviel:
openpose
image
image

lllyasviel:
ok it should be good now

Rkkss:
Canny map in img2img gives bad result. Pixelated + colorshift. Tested at 1024x1024 > 1536x1536 with 0.8 denoise strength.
Scribble, Seg don't have this problem

lllyasviel:
1024x1024 > 1536x1536
just like before, 1024 will require annotator resolution to be also 1024. do not use 512

PR #912, opened 2023-04-19
Update api.py

xdfeng07:
it would be better if the keys in the request stay the same with other parts of the codes, especially for the guessmode and guess_mode


ljleb:
At the moment, the API will accept both `guess_mode` and `guessmode`. See the aliases here:

https://github.com/Mikubill/sd-webui-controlnet/blob/b87cd4c26962b7f287087b57911d960f0626487a/scripts/external_code.py#L164-L169

PR #927, opened 2023-04-20
Update external_code.py

xdfeng07:
if the values of 'mask' is an ndarray, an error will occur for it needs something like any() to discern an ndarray.


lllyasviel:
please commit a "mask is not none"

Issue #43, opened 2023-02-14
combine control net with ultimate sd upscale


Nazzaroth2:
would it be possible to combine controlnet with this extention?
https://github.com/Coyote-A/ultimate-upscale-for-automatic1111 (or would that developer have to integrate the functionality?)

currently it copies the feature-image for every tile that is used in the upscale and thus copies the image over itself. The improved detail coherency from canary or hde might work wonders to allow regaining detailes (especially thinking about hands etc.)

EDIT:
just found the thread talking about general upscaling the mask images. I do think the combination with ultimate upscale would be preferable to general upscaling though, because the individual tile-sizes stay in 512-768px range and thus should work better with the trained models from controlnet, hopefully XP

Danamir:
I concur. 

With segmentation preprocessor, it could identify each segment and avoid rendering faces and hands anywhere. It would really boost the SD upscalers usability !

Preston-L:
With this fantastic combo, one can do incredible style transformations while preserving intricate details!

Lute-4:
+1 :)

https://github.com/Coyote-A/ultimate-upscale-for-automatic1111/issues/36

Danamir:
There must have been some improvements, we can now use controlnet and ultimate sd upscale at the same time, but still with some limitations.

I found the best results by using canny preprocessor, with the annotator resolution set to the sd upscale tile size + padding. There is still a bug, the mask blur is not taken into account so you have to set it to zero and have seams, or have your final image squished.

Here is an example, let's take this 1024x1024 image and upscale it to 1536x1536 by using four 768 tiles. I will be using a 128 padding in SD upscale.

The original image : 
![07102-303429207-photography_,half body portrait, elf woman, cute face, golden braided hair, diaphanous magic armor, fantasy street, detailed env (DPM++ SDE Karras, 16, 6)](https://user-images.githubusercontent.com/15424198/223891928-94b0cd8f-dd9e-462d-8b99-b5f0c1815e29.png)

An upscale without controlnet, and a denoising set  to 0.8 on purpose to illustrate the problem :  
![07154-303429207 0-photography_,half body portrait, elf woman, cute face, golden braided hair, diaphanous magic armor, fantasy street, detailed env (DPM++ 2M Karras, 16, 6)](https://user-images.githubusercontent.com/15424198/223892217-790d6686-687f-4589-939c-1d653fc2168a.png)

[...]

PR #930, opened 2023-04-20
fix controlnet + ultimate sd upscale issue

tfdetang:
try fix https://github.com/Mikubill/sd-webui-controlnet/issues/43

### DIFF
--- a/scripts/controlnet.py
+++ b/scripts/controlnet.py
@@ -852,6 +852,13 @@ class Script(scripts.Script):
                 crop_region = masking.get_crop_region(np.array(mask), p.inpaint_full_res_padding)
                 crop_region = masking.expand_crop_region(crop_region, p.width, p.height, mask.width, mask.height)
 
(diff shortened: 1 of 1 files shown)


lllyasviel:
why ultimate sd upscale has anything to do with mask?

tfdetang:
> why ultimate sd upscale has anything to do with mask?
in ultimate sd upscale, the canva size == mask size;  I think the input image size should == canva size == mask size ;   But when I import a control img, it will not resize to canva size which is unexpected

lllyasviel:
what does ultimate sd upscale use mask for? 

and in fact in controlnet if masks exists, then it will enter some inpainting codes.

tfdetang:
> what does ultimate sd upscale use mask for?
> 
> and in fact in controlnet if masks exists, then it will enter some inpainting codes.


Ultimate sd upscale resize image into target size and cut into tiles and mask/blur them; So I think it is make sense to jump into inpainting codes;

lllyasviel:
already fixed with conflicting codes

PR #939, opened 2023-04-20
Add resize preprocessor

continue-revolution:
The reason why this preprocessor need to exist is that some people may want to use 3rd-party preprocessor, but they may not benefit from the new pixel-perfect feature if this preprocessor does not exist. 

For example, some people may use posex to generate pose and choose None as preprocessor, which might be suboptimal. An incoming new feature of my SAM extension is to use SAM to enhance semantic segmentation, and people may want to use this preprocessor to process the input semantic segmentation image.

### DIFF
--- a/scripts/global_state.py
+++ b/scripts/global_state.py
@@ -51,6 +51,7 @@ cn_preprocessor_modules = {
     "tile_gaussian": tile_gaussian,
     "inpaint": inpaint,
     "invert": invert,
+    "resize": resize,
 }
 
 cn_preprocessor_unloadable = {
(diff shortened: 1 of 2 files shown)


lllyasviel:
pixel-perfect feature and resampling has nothing to do with preprocessor's resize. it is handled inner call and it always influence images even for none canny map. you can try to save a canny map and set none to see

lllyasviel:
you can save a canny map and then use none as preprocessor and enable high-rex fix. you will still see two canny edge maps resampled by controlnet both with one-pixel width lines

continue-revolution:
Sorry for my wasting your time. 

I want to ask why we need to resize the image before it goes into the annotator model. I've noticed that almost all processors with a model will resize the image to a square at the beginning. Is it because the model can only accept/would prefer to accept square images?

lllyasviel:
for 1024 images, canny will produce 1 pixel width lines if annotator resolution is 1024, or 2pixel width lines if resolution is 512. canny ControlNet is trained on images with 1pixel lines.

PR #942, opened 2023-04-21
using both methods to update

lllyasviel:
now we use both methods to update the unit

ignore the line breaks caused buy my ide...


lllyasviel:
it seems that this version works very well. so the solution is to just use both

PR #963, opened 2023-04-22
Move xyz grid suppport and annotator path config

ljleb:
Behold! It may not be obvious right now, but it turns out that this has the potential to transform the earth into a massive black hole and destroy everything if merged into main... 🙄

On a more serious tone, I want to make sure I didn't break anything else again before merging. This changes the modules import configuration, so I need to double check 1 or 2 things.

The point of this change is to allow us to importlib reload annotator_path.py and xyz_grid_support.py, so users can reload the extension without having to completely restart the webui.


ljleb:
Hmm. I tried really hard but it seems that I could not preserve the history of xyz_grid_support.py. If anyone knows how to do it I'd appreciate it.

lllyasviel:
well perhaps we close it if not going to merge - and then reopen when ready

PR #995, opened 2023-04-23
Fix an issue with empty mask string.

ETdoFresh:
### The Issue
When setting up image and mask into nparray... the mask is typically empty. However, when using the API the mask can be an **empty string**, hence controlnet tries to parse the empty string [and causes exception]... So the solution is to check if the incoming mask is a string and is not empty!

### The Fix

Replace

``` python
    if isinstance(result['mask'], str):
        result['mask'] = external_code.to_base64_nparray(result['mask'])
    elif result['mask'] is None:
        result['mask'] = np.zeros_like(result['image'], dtype=np.uint8)
```

with

``` python
    if isinstance(result['mask'], str) and result['mask']:
        result['mask'] = external_code.to_base64_nparray(result['mask'])
    else:
        result['mask'] = np.zeros_like(result['image'], dtype=np.uint8)
```

EDIT: This was not great... look below! 😄 👇

### DIFF
--- a/scripts/controlnet.py
+++ b/scripts/controlnet.py
@@ -119,9 +119,9 @@ def image_dict_from_unit(unit) -> Optional[Dict[str, np.ndarray]]:
     if isinstance(result['image'], str):
         result['image'] = external_code.to_base64_nparray(result['image'])
 
-    if isinstance(result['mask'], str):
+    if isinstance(result['mask'], str) and result['mask']:
         result['mask'] = external_code.to_base64_nparray(result['mask'])
-    elif result['mask'] is None:
+    else:
         result['mask'] = np.zeros_like(result['image'], dtype=np.uint8)
 
     return result
--- a/scripts/controlnet.py
+++ b/scripts/controlnet.py
@@ -117,7 +117,15 @@ def image_dict_from_unit(unit) -> Optional[Dict[str, np.ndarray]]:
     result = {'image': image['image'], 'mask': image['mask']}
 
     if isinstance(result['image'], str):
-        result['image'] = external_code.to_base64_nparray(result['image'])
+        if result['image']:
+            result['image'] = external_code.to_base64_nparray(result['image'])        
+        else:
+            result['image'] = None
+    
+    # If there is no image, ignore the mask
+    if result['image'] is None:
+        result['mask'] = None
+        return None
 
     if isinstance(result['mask'], str):
         if result['mask']:
--- a/scripts/controlnet.py
+++ b/scripts/controlnet.py
@@ -132,35 +132,26 @@ def image_dict_from_any(image) -> Optional[Dict[str, np.ndarray]]:
         # copy to enable modifying the dict and prevent response serialization error
         image = dict(image)
 
-    # copy to enable modifying the dict and prevent response serialization error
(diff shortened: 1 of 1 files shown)


scripts/controlnet.py
```diff
@@ -119,9 +119,9 @@ def image_dict_from_unit(unit) -> Optional[Dict[str, np.ndarray]]:
     if isinstance(result['image'], str):
         result['image'] = external_code.to_base64_nparray(result['image'])
 
-    if isinstance(result['mask'], str):
+    if isinstance(result['mask'], str) and result['mask']:
         result['mask'] = external_code.to_base64_nparray(result['mask'])
-    elif result['mask'] is None:
+    else:
         result['mask'] = np.zeros_like(result['image'], dtype=np.uint8)
```
ljleb, line 125: If the mask is already an np.array, it will get replaced here with zeros. Consider leaving the check for None.

ETdoFresh: ooo yeah.. on it!

ETdoFresh:
Got it... Added nested if statement only if it is a string....
``` python
    if isinstance(result['mask'], str):
        if result['mask']:
            result['mask'] = external_code.to_base64_nparray(result['mask'])
        else:
            result['mask'] = np.zeros_like(result['image'], dtype=np.uint8)    
    elif result['mask'] is None:
        result['mask'] = np.zeros_like(result['image'], dtype=np.uint8)
```

ljleb:
Great. We should probably add the check for `result['image']` as well now that I think about it.

ETdoFresh:
> Great. We should probably add the check for `result['image']` as well now that I think about it.

want me to add it?

ljleb:
If possible! Otherwise I'll do it in the batch PR, no worries.

ETdoFresh:
> If possible! Otherwise I'll do it in a separate PR, no worries.

I just was adding it, and realized the **else part** relies on image. If the image is empty... it'll continue to remain an empty string, and probably just error down the line. soo... errr... maybe this is the best place to cause exception if image is empty?

ljleb:
Well if there is no image but there is a mask, IIUC we should just ignore the mask and return None instead.

ETdoFresh:
Added a bit more logic ignoring both image and mask if image is empty.

``` python
    if isinstance(result['image'], str):
        if result['image']:
            result['image'] = external_code.to_base64_nparray(result['image'])        
        else:
            result['image'] = None
    
    # If there is no image, ignore the mask
    if result['image'] is None:
        result['mask'] = None
        return None

    if isinstance(result['mask'], str):
        if result['mask']:
            result['mask'] = external_code.to_base64_nparray(result['mask'])
        else:
            result['mask'] = np.zeros_like(result['image'], dtype=np.uint8)    
    elif result['mask'] is None:
        result['mask'] = np.zeros_like(result['image'], dtype=np.uint8)
```

lllyasviel:
please resolve conflicts and will merge

ETdoFresh:
Not yet resolved... I'll message back when ready.

ETdoFresh:
OK, that should do it!

BuffMcBigHuge:
I was able to temporarily resolve this issue by setting `mask: null` in the API request.

PR #998, opened 2023-04-23
simplify ui

lllyasviel:


ljleb:
One problem with this: I want to be able to enable loopback even with units that are in Upload mode.

ljleb:
The renaming is fine, but it makes some useful usage less accessible.

lllyasviel:
wait, what does loopback do? does it loop the batch files?

ljleb:
loopback takes the first image of the last generation during a batch and replaces the input image to the units that has it toggled on.

ljleb:
TL;DR: TemporalNet needs this to work.

lllyasviel:
what is "the first output image of the last iteration during a batch", the batch in a1111 or batch in new new "Batch"?

ljleb:
the img2img batch tab, as well as the custom hijack batch tab in batch_hijack.py. It is not the same as batch count + batch size in the ui settings.

batch count + batch size in the settings = 1 batch tab iteration.

lllyasviel:
so it actually does not loop the files in "Batch"?

lllyasviel:
if loopback is on, then i generate some images, my controlnet image will be replaced with one of the generated result?

ljleb:
> if loopback is on, then i generate some images, my controlnet image will be replaced with one of the generated result?

Yes, but only after the first batch image. (so the second iteration of our custom batch will be replaced) It will not replace the initial input image of any unit after processing. So if you press generate multiple times, it only takes effect during batch processing.

lllyasviel:
when i see the new Batch Tab, my feeling is that:
1. if loopback is off, "Batch Tab" randomly read images in a folder as control image.
2. if loopback is on, "Batch Tab" read images in a folder in a order and loop it.

lllyasviel:
if that is not the case, then the ui naming should be revised since it mislead me on time and will confuse people like me

ljleb:
It is instead like this:

- loopback off: read images in the specified folder in the same order as other units
- loopback on:
    - iteration 1: read the first image of the folder (or upload image if unit is in simple mode)
    - iteration n > 1: use the last generated image and pass it inside the preprocessor again (if any)

lllyasviel:
what about "automatically send output to control image upload"?

ljleb:
I'm okay with that. I've heard the term "loopback" being thrown around a lot for this concept, maybe we could add a tooltip instead? Since it clutters the interface a little bit.

lllyasviel:
image

lllyasviel:
what about this

ljleb:
Why not. We don't need all the checkboxes at the same place I guess 👍

lllyasviel:
https://github.com/Mikubill/sd-webui-controlnet/pull/999

ljleb:
I think I would use "automatically "send last generated image into this control unit" or something that says "this control unit" for more clarity.

lllyasviel:
but i tested it and it does not seem to work

lllyasviel:
i check it and but it does not send result to controlnet

ljleb:
Can you show what you see and what you expect to see? I ran multiple situations and couldn't get it to fail. Make sure at least one of the units is in batch mode.

lllyasviel:
i do not use batchmode, just use single image

lllyasviel:
but if i need batchmode to use it, why not put it in the "Batch Tab"

lllyasviel:
https://github.com/Mikubill/sd-webui-controlnet/pull/1000

ljleb:
It does not move anything in the user interface. It only takes effect during batch processing. However, I want to be able to upload the first image and then let the unit receive the output of each iteration after.

See how temporalnet is used here: https://huggingface.co/CiaraRowles/TemporalNet/blob/d47bd9b63c1717bdfecda4e75653dcf0d7219f85/temporalvideo.py#L98

I want to be able to do this:

image
image

So unit 1 is in upload mode, but has loopback, whereas unit 2 is in batch mode but NOT in loopback mode.

lllyasviel:
cool then perhaps just in this shape

PR #1003, opened 2023-04-24
[UI Improvement] Add hover tooltip for ToolButton

huchenlei:
I found the controlnet emoji buttons a little bit confusing. The user might need to try really hard to guess what each button is for. This PR adds a hover tooltip on each `ToolButton` which appears after the user hovers on the button for a while. 

There is a simpler way to display hover text by adding `title` attribute on `button` element[[link]](https://stackoverflow.com/questions/45456543/make-text-show-up-on-hover-over-button).  However, I did not found a way in Gradio to either directly add a custom attribute on button element, or trigger a piece of JS code that add custom attribute to the element on document load. 

So, I have to take a far route which is pure CSS based implementation.
Tooltip

Note:
There is a undesired side-effect on the UI. Gradio has following CSS rule on non-absolute elements. 
```
div.svelte-1oo81b7>*:not(.absolute) {
    border-radius: 0!important;
}
```
This will make the bottom-right and bottom-left coners of the button not rounded. 

If you think this is a big issue, I can workround on CSS to fix this UI side effect.

### DIFF
--- a/javascript/hints.js
+++ b/javascript/hints.js
@@ -5,7 +5,7 @@ titles = {
     '💥': 'Run preprocessor',
     '📝': 'Open new canvas',
     '📷': 'Enable webcam',
-    '⇄': 'Mirrow webcam',
+    '⇄': 'Mirror webcam',
 };
 
 onUiUpdate(function(){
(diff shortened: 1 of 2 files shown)


huchenlei:
Note: 
`elem_classes` seem to be broken in earlier versions of Gradio, so probably we need to guard the tooltip implementation behind `gradio_compat` flag?

lllyasviel:
it seems add css is a relatively dangerous thing and some other plugins break a1111 with it constantly. we need more test before merge

PR #1006, opened 2023-04-24
Fixed issue where Nontype causes a crash

jmsether:
On some plugins this crashes on object type Nonetype. Figured because the function can return type none it would be a good idea to return type none if we have a variable of type none. There are a lot of nones in this post. Does that make it holly? (sorry for the poor joke, I have been starring at python code all day)

image


lllyasviel:
@ljleb this pull failed github tests. please fix or revert.

jmsether:
Hmmm. Not sure what went wrong here. find_cn_script returns(or can return None) right?

ljleb:
Didn't see the additional colon.

PR #1013, opened 2023-04-24
ignore extra args from api and fix param count

ljleb:


ljleb:
not sure why tests pass in main, as the flattened args are missing the new control parameter. #678 may be related.

Issue #1037, opened 2023-04-25
[Bug]: What is the API keyword for selecting "control mode" (guess mode) in 1.1?


Jot-De:
### Is there an existing issue for this?

- [X] I have searched the existing issues and checked the recent builds/commits of both this extension and the webui

### What happened?

The API keyword guessmode no longer has any effect in ControlNet 1.1?
What is the correct keyword in API to change control mode (guess mode) to balanced / prompt / control?

### Steps to reproduce the problem

```
# prepare payload for POST
method payload = {  
  "init_images": [img], 
  "denoising_strength": 1,
   "prompt": "a woman",  
 "sampler_name": "DDIM",
   "steps": 30,  
 "cfg_scale": 7.5, 
  "negative_prompt": "poor quality",  
 "include_init_images": True,  
 "seed":23,   
 "batch_size":1,  
 "alwayson_scripts": {   
  "controlnet": {   
      "args": [  
         {   
          "model": "control_hed-fp16 [13fee50b]",  
           "enabled":True, 
	   "input_image":[img],  
           "module": "softedge_hed",  
           "weight": 2.0,
             "processor_res": 512,  
           "guidance_end": 1.0,  
           "guessmode": True
           }]   
}
 }
 }
 # send request 
response = requests.post(url=f'{url}/sdapi/v1/txt2img', json=payload) 
r = response.json() 
output = Image.open(io.BytesIO(base64.b64decode(r['images'][0].split(",",1)[0]))) 
output

```

### What should have happened?

guessmode OR any other proper keyword which I don't know should change guess mode.

### Commit where the problem happens

webui:  22bcc7b
controlnet:  [a07f6e8](https://github.com/Mikubill/sd-webui-controlnet/commit/a07f6e8a1a4d4d6ced366aec30fd8244f1fde26d)

### Command Line Arguments

```Shell
--api
```

### Console logs

```Shell
Loading model from cache: control_hed-fp16 [13fee50b]
Loading preprocessor: hed
preprocessor resolution = 512
Data shape for DDIM sampling is (1, 4, 64, 64), eta 0.0
Running DDIM Sampling with 31 timesteps
DDIM Sampler: 100%|████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 31/31 [00:06<00:00,  4.61it/s] 
Total progress: 93it [10:08,  4.61it/s]
```

joaxkal:
[...]

PR #1038, opened 2023-04-25
convert control mode

ljleb:
related: #1037


ljleb:
`"guess_mode"` now maps to `ControlMode.CONTROL` and takes precedence over existing values.

PR #1040, opened 2023-04-26
add api module alias

ljleb:


ljleb:
What is the next version after 109? We passed from 9 to 101 last time so the pattern isn't clear to me.

lllyasviel:
it should be 110

PR #1073, opened 2023-04-28
[Unittest] Add unittest on openpose face/hand detection functions

huchenlei:
This PR adds unittests on `faceDetect` and `handDetect` functions. 

I am planning to do some refactorings to those 2 functions, so adding some unittests to make sure that I do not alter the behaviour.


huchenlei:
Can someone explain why existing tests start to fail when this new test is added? It seems like tests cannot correctly locate the `annotator` module?

PR #1077, opened 2023-04-28
[Refactor] Adjust body detection result representation for easier JSON serialization

huchenlei:
Previously the intermediant representation before drawing pose on canvas was not well aligned with [openpose JSON output format](https://github.com/CMU-Perceptual-Computing-Lab/openpose/blob/master/doc/02_output.md#face-output-format).

This PR does a refactor so that we have a cleaner datastructure for poses.

In the old format, we were using `candidates` and `subset` to represent bodies in the image, where:
- `candidates`: an array of 2D points
- `subset`: an 2D array with each row representing a body, each cell reference a 2D points in the candidates array

Now we have a clean `BodyResult` named tuple to represent this. The `BodyResult` directly stores all 2D points so that the caller does not need to do the index lookup work.

The correctness of this refactoring is verified by unittests added in previous PRs.

The next few PRs will add JSON encode / download feature.

Note:
This CL mostly preserves the original logic of the code, but I have a question:
Why do we normalize the x, y values of the keypoints in the detection function, and do the reverse in the draw function?
Is the normalization necessary in the very first place?


annotator/openpose/util.py
```diff
@@ -65,149 +67,224 @@ def transfer(model, model_weights):
...
 
+    Note:
+        The function expects the x and y coordinates of the keypoints to be normalized between 0 and 1.
+    """
+    H, W, C = canvas.shape
     stickwidth = 4
 
-    limbSeq = [[2, 3], [2, 6], [3, 4], [4, 5], [6, 7], [7, 8], [2, 9], [9, 10], \
-               [10, 11], [2, 12], [12, 13], [13, 14], [2, 1], [1, 15], [15, 17], \
-               [1, 16], [16, 18], [3, 17], [6, 18]]
```
huchenlei, line 77: Interestingly even we only use the first 17 elements in `limbSeq`, there are 19 elements in the list. This PR removes the last 2 items in the list.

annotator/openpose/body.py
```diff
@@ -7,10 +7,23 @@
...
 from torchvision import transforms
+from typing import NamedTuple, List
 
 from . import util
 from .model import bodypose_model
 
+class Keypoint(NamedTuple):
+    x: float
+    y: float
+    score: float = 1.0
```
huchenlei, line 18: `score` and `id` are not used anywhere else in the code. Do we want to keep them? Is `score` here the same concept of `confidence`?

annotator/openpose/util.py
```diff
@@ -65,149 +67,224 @@ def transfer(model, model_weights):
...
     colors = [[255, 0, 0], [255, 85, 0], [255, 170, 0], [255, 255, 0], [170, 255, 0], [85, 255, 0], [0, 255, 0], \
               [0, 255, 85], [0, 255, 170], [0, 255, 255], [0, 170, 255], [0, 85, 255], [0, 0, 255], [85, 0, 255], \
               [170, 0, 255], [255, 0, 255], [255, 0, 170], [255, 0, 85]]
 
-    for i in range(17):
-        for n in range(len(subset)):
-            index = subset[n][np.array(limbSeq[i]) - 1]
-            if -1 in index:
-                continue
-            Y = candidate[index.astype(int), 0] * float(W)
```
huchenlei, line 88: I believe here we have X and Y reversed somehow. Do we have a reason for this naming here? The naming is left untouched in this PR.

annotator/openpose/util.py
huchenlei: Ref:
https://github.com/CMU-Perceptual-Computing-Lab/openpose/blob/master/.github/media/keypoints_pose_18.png

PR #1085, opened 2023-04-29
fix hints (toole tip) for webui when controlnet is enabled

w-e-w:
the hints (toole tip) used in controlnet overrides the hints used in webui
which extension enabled, the [hints in webui](https://github.com/AUTOMATIC1111/stable-diffusion-webui/blob/master/javascript/hints.js) is broken
this fixes it

I'm not a web developer, so I'm not really familiar with JavaScript, I suspect they might be a more elegant way of fixing this.

### DIFF
--- a/javascript/hints.js
+++ b/javascript/hints.js
@@ -1,18 +1,17 @@
-// mouseover tooltips for various UI elements
-titles_controlnet = {
-    '🔄': 'Refresh',
-    '\u2934': 'Send dimensions to stable diffusion',
-    '💥': 'Run preprocessor',
-    '📝': 'Open new canvas',
-    '📷': 'Enable webcam',
-    '⇄': 'Mirror webcam',
-};
-
-onUiUpdate(function(){
-	gradioApp().querySelectorAll('.cnet-toolbutton').forEach(function(button){
-		tooltip = titles_controlnet[button.textContent];
-		if(tooltip){
-			button.title = tooltip;
-		}
-	})
+onUiUpdate(function () {
+    // mouseover tooltips for various UI elements
+    const titles = {
+        '🔄': 'Refresh',
+        '\u2934': 'Send dimensions to stable diffusion',
+        '💥': 'Run preprocessor',
+        '📝': 'Open new canvas',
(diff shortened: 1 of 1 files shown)


huchenlei:
My bad. I just copied the implementation from webui hint.js. The best way to fix this is probably move `hints` to function local scope. Also `title` need to be declared as local variable (currently it is overwriting the global `title`).

huchenlei:
The more elegant fix:
```
onUiUpdate(function(){
        // mouseover tooltips for various UI elements
        const titles = {
            '🔄': 'Refresh',
            '\u2934': 'Send dimensions to stable diffusion',
            '💥': 'Run preprocessor',
            '📝': 'Open new canvas',
            '📷': 'Enable webcam',
            '⇄': 'Mirror webcam',
        };
	gradioApp().querySelectorAll('.cnet-toolbutton').forEach(function(button){
		const tooltip = titles[button.textContent];
		if(tooltip){
			button.title = tooltip;
		}
	})
});
```

w-e-w:
another methond of fixing this can be
```js
titles_controlnet = {
    '🔄': 'Refresh',
    '\u2934': 'Send dimensions to stable diffusion',
    '💥': 'Run preprocessor',
    '📝': 'Open new canvas',
    '📷': 'Enable webcam',
    '⇄': 'Mirror webcam',
};

titles = Object.assign(titles, titles_controlnet);
```

essentially this merges the hint from controlnet to the dictionary in the base webui

the upside of this method is possible negligible performance Improvement
the downside is that if anything were to change on the webub side
it might affect here

huchenlei:
We probably don't want to mess with the web-ui hints, as the hints for icons are specific to controlnet. The same icon can have different meaning when used elsewhere.

w-e-w:
I update it to use your method

huchenlei:
Thanks for spot and fixing this!

w-e-w:
I spotted it because I was editing tooltip over at webui and was very confused that it wasn't showing

lllyasviel:
is it safe to merge now?

Issue #1101, opened 2023-05-01
[Unittest] web_api/text2img_test.py always set `module="none"`


huchenlei:
Our current unittest sets `module="none"` in all test cases. This leaves the code path of calling preprocessors uncovered.

huchenlei:
Preprocessor `clip` seems to be failing with following exception during local testing:
```
API error: POST: http://localhost:7860/sdapi/v1/txt2img {'error': 'RuntimeError', 'detail': '', 'body': '', 'errors': 'Expected 3D (unbatched) or 4D (batched) input to conv2d, but got input of size: [257, 1024]'}
Traceback (most recent call last):
  File "D:\stable-diffusion-webui\venv\lib\site-packages\anyio\streams\memory.py", line 94, in receive
    return self.receive_nowait()
  File "D:\stable-diffusion-webui\venv\lib\site-packages\anyio\streams\memory.py", line 89, in receive_nowait
    raise WouldBlock
anyio.WouldBlock

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "D:\stable-diffusion-webui\venv\lib\site-packages\starlette\middleware\base.py", line 78, in call_next
    message = await recv_stream.receive()
  File "D:\stable-diffusion-webui\venv\lib\site-packages\anyio\streams\memory.py", line 114, in receive
    raise EndOfStream
anyio.EndOfStream

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "D:\stable-diffusion-webui\modules\api\api.py", line 145, in exception_handling
    return await call_next(request)
  File "D:\stable-diffusion-webui\venv\lib\site-packages\starlette\middleware\base.py", line 84, in call_next
    raise app_exc
  File "D:\stable-diffusion-webui\venv\lib\site-packages\starlette\middleware\base.py", line 70, in coro
    await self.app(scope, receive_or_disconnect, send_no_error)
  File "D:\stable-diffusion-webui\venv\lib\site-packages\starlette\middleware\base.py", line 108, in __call__
    response = await self.dispatch_func(request, call_next)
  File "D:\stable-diffusion-webui\modules\api\api.py", line 110, in log_and_time
    res: Response = await call_next(req)
  File "D:\stable-diffusion-webui\venv\lib\site-packages\starlette\middleware\base.py", line 84, in call_next
    raise app_exc
[...]

PR #1102, opened 2023-05-01
[Unittest] Add txt2img test with every preprocessors

huchenlei:
This PR addresses #1101.

### DIFF
--- a/tests/utils.py
+++ b/tests/utils.py
@@ -3,6 +3,7 @@ from base64 import b64encode
 
 import requests
 
+BASE_URL = "http://localhost:7860"
 
 def setup_test_env():
     ext_root = os.path.dirname(os.path.dirname(os.path.realpath(__file__)))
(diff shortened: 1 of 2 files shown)


huchenlei:
Runner says:
```
Error: The operation was canceled.
##[debug]System.OperationCanceledException: The operation was canceled.
```

But I did not cancel it manually.

lllyasviel:
maybe many preprocessor need gpu

lllyasviel:
wait it seems that new a1111 break all github tests

huchenlei:
Seems like we don't have necessary resources to run every preprocessor in github CI. I will make the test running locally presubmit.

lllyasviel:
what  about only test openpose_full and depth

lllyasviel:
wow the test passed, merge here?

huchenlei:
Sure we can add more coverage if we fell necessary later.

PR #1131, opened 2023-05-02
Add back type annotation on BodyResult

huchenlei:
The pipe operation on type annotation is introduced in Python 3.10. See [PEP 604](https://peps.python.org/pep-0604/).

User ran into issue when running with 3.8. 

Add back type annotation as it is a problem with the user's setup. 

Note:
I am not pumping up version in this PR as type annotation does not affect code execution.

### DIFF
--- a/annotator/openpose/body.py
+++ b/annotator/openpose/body.py
@@ -7,7 +7,7 @@ import matplotlib.pyplot as plt
 import matplotlib
 import torch
 from torchvision import transforms
-from typing import NamedTuple, List
+from typing import NamedTuple, List, Union
 
 from . import util
 from .model import bodypose_model
@@ -20,7 +20,12 @@ class Keypoint(NamedTuple):
 
 
 class BodyResult(NamedTuple):
-    keypoints: List[Keypoint | None]
+    # Note: Using `Union` instead of `|` operator as the ladder is a Python
+    # 3.10 feature.
+    # Annotator code should be Python 3.8 Compatible, as controlnet repo uses
+    # Python 3.8 environment.
+    # https://github.com/lllyasviel/ControlNet/blob/d3284fcd0972c510635a4f5abe2eeb71dc0de524/environment.yaml#L6
+    keypoints: List[Union[Keypoint, None]]
     total_score: float
     total_parts: int
 
--- a/annotator/openpose/body.py
+++ b/annotator/openpose/body.py
@@ -20,7 +20,7 @@ class Keypoint(NamedTuple):
 
 
 class BodyResult(NamedTuple):
-    keypoints: List
+    keypoints: List[Keypoint | None]
     total_score: float
     total_parts: int
 


ljleb:
Alternatively we can use the Union type. It is less concise but will not prevent older python to parse the code. Though, I agree that, as the webui should be ran with 3.10.6, it may be acceptable to just use the pipe.

lllyasviel:
or can we just push a typing.Union

lllyasviel:
image
from typing import List, Union

class Keypoint:
    def __init__(self, x: float, y: float):
        self.x = x
        self.y = y

keypoints: List[Union[Keypoint, None]] = [Keypoint(0.1, 0.2), None, Keypoint(0.3, 0.4)]

huchenlei:
Since the WebUI is not meant to be run in a legacy version of Python, we should assume that all features since 3.10 are available. A legacy version of Python is not expected to run WebUI and its extensions correctly. 

Since the new pipe syntax introduce less noise, I still insists that we use the new pipe syntax on union types.

lllyasviel:
well we consider that the openpose codes are copied from controlnet and controlnet's env is copied from stability, 

controlnet 3.8.5 https://github.com/lllyasviel/ControlNet/blob/d3284fcd0972c510635a4f5abe2eeb71dc0de524/environment.yaml#L6

stability 3.8.5 https://github.com/Stability-AI/stablediffusion/blob/cf1d67a6fd5ea1aa600c4df58e5b47da45f6bdbf/environment.yaml#L6

if this becomes 3.10 then it cannot be copied back to controlnet, which may cause some offsets between community training or some else (if they really copy it back to some training repo)

ljleb:
The webui technically requires 3.10.6, or at least that's what's advertised in the installation guide wiki. Being able to copy the code to 3.8.5 code bases means the extension has to support both versions instead of just one or the other. The changes from 3.8.5 to 3.10.6 are not dramatic AFAIK but this would imply we need to avoid using mutually exclusive language features.

continue-revolution:
@huchenlei 
Why not just use typing.Union instead? Updating Python version means updating the whole virtual environment which is very time consuming. It is not worth asking all ControlNet users to update their whole virtual environment just for a seemingly more beautiful type annotation. This will influence thousands of users. But using typing.Union instead of ｜may only take you 10 minutes and just make your code slightly less clean but completely acceptable.

huchenlei:
> well we consider that the openpose codes are copied from controlnet and controlnet's env is copied from stability,
> 
> controlnet 3.8.5 https://github.com/lllyasviel/ControlNet/blob/d3284fcd0972c510635a4f5abe2eeb71dc0de524/environment.yaml#L6
> 
> stability 3.8.5 https://github.com/Stability-AI/stablediffusion/blob/cf1d67a6fd5ea1aa600c4df58e5b47da45f6bdbf/environment.yaml#L6
> 
> if this becomes 3.10 then it cannot be copied back to controlnet, which may cause some offsets between community training or some else (if they really copy it back to some training repo)

Makes sence. So in annotator related code we should use Python 3.8 compatible features, and in WebUI related code we can use Python 3.10 features. Am I understanding it correctly?

lllyasviel:
well it seems that webui is also calling https://github.com/Stability-AI/stablediffusion 's ldm so that webui is actually using 3.8.5 codes, so it should be safer to see it as 3.8. Some other standalone XXXX trainer may want to call SD2's ldm for training and perhaps they want to also call some funcs in extensions.

lllyasviel:
seems that the reason that SD2 is not using 3.10 is related to a problem that pytorch_lightening's 3.10 does not support a needed xformer feature in 3.8

continue-revolution:
https://github.com/Mikubill/sd-webui-controlnet/blob/main/annotator/openpose/util.py#L127
https://github.com/Mikubill/sd-webui-controlnet/blob/main/annotator/openpose/util.py#L172
https://github.com/Mikubill/sd-webui-controlnet/blob/main/annotator/openpose/util.py#L296
should also be updated.

lllyasviel:
wait do we have so many 3.10 only codes?

continue-revolution:
yes, https://github.com/Mikubill/sd-webui-controlnet/blob/main/annotator/openpose/__init__.py#L34 also 3.10. I'm still testing how many such stuff are present in this repo

lllyasviel:
wait but if those are just hints rather than real codes perhaps we just ignore none

continue-revolution:
They are just hints. I can shoot a PR in a couple minutes to bypass this issue, after openpose preview works on my side.

PR #1133, opened 2023-05-02
Replace pipe operator on type annotation with 3.8 compatible syntax

huchenlei:


continue-revolution:
Optional is also acceptable which might be more clean if you want

PR #1154, opened 2023-05-05
Remove deprecated api

ljleb:
Remove the deprecated inference API routes. Discussion: #1132

### DIFF
--- a/scripts/api.py
+++ b/scripts/api.py
@@ -1,12 +1,7 @@
-from typing import Union
-
 import numpy as np
 from fastapi import FastAPI, Body
 from fastapi.exceptions import HTTPException
 from PIL import Image
-import copy
-import pydantic
-import sys
 
 import gradio as gr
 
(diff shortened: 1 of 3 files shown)


ljleb:
raised the api version as well since this is a breaking change.

lllyasviel:
will this break ci test? do we remove old tests?

ljleb:
no it does not break ci tests, as I could not find any test for the deprecated api routes.

ljleb:
The failing pipeline above is because I manually cancelled it.

lllyasviel:
but what are those test_deprecated_xxx?

lllyasviel:
what is TestDeprecatedTxt2ImgWorking?

ljleb:
It's for flattened args. We used to be able to pass all args as a flattened list instead of a list of dicts. We can remove this feature as well, we do not need it anymore and I don't think anyone is using it in this case.

lllyasviel:
do it

PR #1155, opened 2023-05-05
Remove flat args

ljleb:
Removes flattened ui args. As we added and removed options this last month, and no one complained about the change, we can be more or less certain no one is using it.


lllyasviel:
i will merge a pr to remove guess mode after this pr is merged

ljleb:
didn't have time to update the version. I guess it doesn't matter as your PR is close behind

PR #1156, opened 2023-05-05
Completely remove deprecated guess mode

lllyasviel:


ljleb:
It is possible this breaks some extensions or external code. Are you sure this is a good idea? I think it is a better idea to let everybody know before making this breaking change.

lllyasviel:
we already have **kwargs so it wont break things

ljleb:
See for example: https://github.com/mix1009/sdwebuiapi/blob/650a2c93269e7e2931bee6e3f8134d7d8e536f89/webuiapi/webuiapi.py#L692

ljleb:
We should tell people in one way or another, either though a discussion or a warning in console.

lllyasviel:
good. added print

PR #1166, opened 2023-05-05
fetch sliders_config by `/controlnet/module_list`

zombieyang:
I'm contributing to the [Auto-Photoshop-StableDiffusion-Plugin](https://github.com/AbdullahAlfaraj/Auto-Photoshop-StableDiffusion-Plugin)
When I was implementing the ThresholdA/B in it, I found that all the sliders config is hardcoded in sd-webui-controlnet.
I think it is better to expose them in the API, so external tools like [Auto-Photoshop-StableDiffusion-Plugin](https://github.com/AbdullahAlfaraj/Auto-Photoshop-StableDiffusion-Plugin) or plugin in Blender etc. could build ThresholdA/B sliders without hardcoding the settings of sliders in their code.

### DIFF
--- a/scripts/api.py
+++ b/scripts/api.py
@@ -9,6 +9,7 @@ from modules.api.models import *
 from modules.api import api
 
 from scripts import external_code, global_state
+from scripts.processor import preprocessor_sliders_config
 
 def encode_to_base64(image):
     if type(image) is str:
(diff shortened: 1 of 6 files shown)


lllyasviel:
wow, will test

scripts/api.py
```diff
@@ -37,9 +38,25 @@ async def model_list():
...
+                }
+            else:
+                _module_detail[module] = {
+                    "sliders": []
+                }
+        
+        return {
+            "module_list": _output_list,
+            "module_detail": _module_detail
+        }
```
ljleb, line 59: this looks like a good contribution. you think it would make sense to expose the module_detail values through the external_code api as well?

lllyasviel:
unfornately this seems to break pixel perfect toggle's hidding preprocessor resolution

lllyasviel:
seems that we need to refine the logic a bit

scripts/api.py
zombieyang: I had never used the external_code api before...Let me have a look.

zombieyang:
@ljleb @lllyasviel Hello, I have fixed my bugs (I submitted wrong file this morning) and add the function in `external_code.py`.

I also add a simple test in `txt2img_test/test_call_with_preprocessors`

however some other test is failed in the github action of my repo but it seems like not my business.

lllyasviel:
let me take a look soon

PR #1174, opened 2023-05-06
fix ci

lllyasviel:


lllyasviel:
F

lllyasviel:
what happened?

ljleb:
my guess is either runners are stateful, or there is a race condition between push and pull_request events

PR #1175, opened 2023-05-06
fix ci

lllyasviel:


ljleb:
if your intention was to reduce the number of commits, you may want to consider using squash commits when merging. I think we should always squash commits on merge, as sometimes the history of different branches becomes intertwined in main

PR #1206, opened 2023-05-08
[fix]Adding elem_id to all the gradio component of controlnet

guming3d:
Adding elem_id to all the gradio component of controlnet, we have a extension which use jspath to grap the controlnet component value in DOM, need to set the elem_id otherwise every release will got different jspath.

### DIFF
--- a/scripts/controlnet.py
+++ b/scripts/controlnet.py
@@ -214,10 +214,10 @@ class Script(scripts.Script):
         self.unloadable = global_state.cn_preprocessor_unloadable
         self.input_image = None
         self.latest_model_hash = ""
-        self.txt2img_w_slider = gr.Slider()
-        self.txt2img_h_slider = gr.Slider()
-        self.img2img_w_slider = gr.Slider()
-        self.img2img_h_slider = gr.Slider()
+        self.txt2img_w_slider = gr.Slider(elem_id='txt2img_w_slider')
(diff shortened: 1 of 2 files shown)


lllyasviel:
oh and biggest problem.
when multi controlnet is turned on, if user has N controlnets, you will have N dom element with exactly same elem_id for each name.
The solution is to use prefix. See also how we handle preview_check_elem_id.
This could prevent the pr from merge.

guming3d:
@lllyasviel  what does "when multi controlnet is turned on" mean, can we install multiple controlnet extensions at same time?

guming3d:
BTW, our method to us javascript to grab the controlnet component value is too hack, and easy to fail if jsPath is changed by release. Is there any plan to save the component value to a local js file? Thanks a lot

guming3d:
I have updated the elem_id to use prefix like preview_check_elem_id in the PR, please help to check again

lllyasviel:
> @lllyasviel what does "when multi controlnet is turned on" mean, can we install multiple controlnet extensions at same time?

u can enable multi controlnet in setting

huchenlei:
To avoid conflicted element id, you can set `elem_classes` instead, and using JS query selector to select all elements with certain class, e.g. `document.querySelector(".cnet-canvas-width")`.

PR #1232, opened 2023-05-13
first step in supporting more models

lllyasviel:
This is the first step in supporting more models
this is already tested and can be merged whenever 
but will merge after extensive tests


lllyasviel:
since 1.1.150 is a very robust version and has holded for 4 days without reproducible bug resoprts - we do not want to break things

this pull will be more carefully tested.

if anyone want to help testing then u are welcome

note that we must have this merged before adding more models so this pull is actually inevitable, sooner or later

PR #1273, opened 2023-05-14
Style Fidelity Slider

lllyasviel:
Since different people have different preferences, we will add a Style Fidelity Slider slider since 1.1.168
image
This slider only work in "balanced" mode
higher value mean more fidelity in style, but also more risk to collapse
lower value mean less fidelity in style, but also more robust


vonvanzu:
I just updated to 'ControlNet v1.1.173' in 'SD webui AUTOMATIC1111" in google colab but the slider doesn't show.

2blackbar:
Is it possible to uise this slider with all controlnet models ?

PR #1278, opened 2023-05-15
add preprocessor reference_adain and adain_attn

lllyasviel:
V1.1.171 adds two new reference preprocessors:

**reference_adain**
AdaIn (Adaptive Instance Normalization) from
Arbitrary Style Transfer in Real-time with Adaptive Instance Normalization
https://arxiv.org/abs/1703.06868

**reference_adain+attn**
AdaIn (Adaptive Instance Normalization) + Attention link same as "reference_only"

### Comparison
source (midjourney v5, https://twitter.com/kajikent/status/1654409097041817601)

meta

woman in street, masterpiece, best quality,
Negative prompt: lowres, bad anatomy, bad hands, text, error, missing fingers, extra digit, fewer digits, cropped, worst quality, low quality, normal quality, jpeg artifacts, signature, watermark, username, blurry
Steps: 20, Sampler: Euler a, CFG scale: 7, Seed: 12345, Size: 768x512, Model hash: c0d1994c73, Model: realisticVisionV20_v20, Version: v1.2.0, ControlNet 0: "preprocessor: reference_????, model: None, weight: 1, starting/ending: (0, 1), resize mode: Crop and Resize, pixel perfect: True, control mode: Balanced, preprocessor params: (64, 0.5, 64)"

reference_only:
image

reference_adain:
image

reference_adain+attn:
image

without CN (all controlnets disabled):
image

### DIFF
--- a/scripts/hook.py
+++ b/scripts/hook.py
@@ -30,9 +30,9 @@ class ControlModelType(Enum):
 
 
 # Written by Lvmin
-class AttentionAutoMachine(Enum):
+class AutoMachine(Enum):
     """
-    Lvmin's algorithm for Attention AutoMachine States.
+    Lvmin's algorithm for Attention/AdaIn AutoMachine States.
     """
 
     Read = "Read"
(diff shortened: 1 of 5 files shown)


scripts/hook.py
```diff
@@ -440,20 +455,66 @@ def hacked_basic_transformer_inner_forward(self, x, context=None):
             x = self.ff(self.norm3(x)) + x
             return x
 
+        def hacked_group_norm_forward(self, *args, **kwargs):
```
okotaku, line 458: The function name "hacked_group_norm_forward" seems to suggest hacking the group_norm. But it actually seems to be hacking the forward functions of the "ResBlock" and "SpatialTransformer" blocks. 
Am I correct in understanding that it hacks the forward functions of the blocks, rather than the group norm?

lllyasviel: The intention is to hack any features before feed it to group norm - but it seems not important as we can apply adain to any layers.

okotaku: I understand, thank you!

PR #1286, opened 2023-05-15
no clear model cache when sd model changes

dennissheng:
When sd model changes, control model cache will be cleared. Maybe we can cache muliti sd model?  Just need to change the format of the cache key, like "sd_model + control_model".


lllyasviel:
wait - is this fixing error or optimizing mem?
is the current code wrong and can we trigger some bugs with current code logic？

lllyasviel:
ah i know - this only influence diff controlnet right? 
but considering we are going to end the support for diff controls, perhaps not going to add complexity for diff controlnet

dennissheng:
@lllyasviel yes, this is an optimization of decreasing loading time when switching stable diffusion model.  In my test, this would decrease several seconds.

Issue #1036, opened 2023-04-25
[Bug]: Pre-processors seem broken


nperez:
### Is there an existing issue for this?

- [X] I have searched the existing issues and checked the recent builds/commits of both this extension and the webui

### What happened?

Not entirely sure where I messed up, but my pre-processors simply return black or all grey images or greyscale interference patterns like this: 
  
![tmpx7nyz_r5](https://user-images.githubusercontent.com/75055/234334539-896c965a-801b-4d2e-a794-299183cef357.png) 

or like this:  
  
![image](https://user-images.githubusercontent.com/75055/234336664-fe1e9f34-0af9-45c8-9d53-864c05523b7e.png)


I am currently running webui at `22bcc7be428`, and the controlnet extension at `a07f6e8a1a` on Debian. My setup is a bit funky though. I used conda to make sure I use python 3.10 (so all of the wheels install nicely in the venv of webui), and have modified the dependencies of webui to run torch v2 (torch==2.0.0.+cu118 torchvision==0.15.1+cu118). Then I also downloaded the controlnet v1.1 models and was using them.

Everything was fine earlier this week. But when I spun up the webui today to play, is when I noticed the pre-processor behavior.

Here's what I've tried so far:

1. reflog spelunking on the controlnet extension repo and going back two or three refs in time
2. git clean -dxf in webui and controlnet extension repos (which also wipes out venv in webui and thus deps were reinstalled)
3. double checking my conda env is correct and activated (it is afaict)
4. Tried different images just in case I was feeding it something it didn't like

Pretty sure I am holding it wrong, but I'd appreciate some guidance.

### Steps to reproduce the problem

1. Add image
2. Allow Preview
3. Select pre-processor (softedge_hed)
4. Click the explody button
5. ???
6. Profit

### What should have happened?

It should behave how it behaved previously.

### Commit where the problem happens

webui: `22bcc7be428`
controlnet extension: `a07f6e8a1a` 

And you can see what refs I tried out again here:

$ cat .git/logs/refs/heads/main 
0000000000000000000000000000000000000000 241c05f8c9d3c5abe637187e3c4bb46f17447029 nperez@cpan.org 1680787793 +0200       clone: from https://github.com/Mikubill/sd-webui-controlnet.git
[...]

PR #1302, opened 2023-05-16
Prevent double print of version

catboxanon:
Prevents the version number printing twice in the console (example of current behavior can be seen [here](https://github.com/Mikubill/sd-webui-controlnet/issues/1036)). I don't believe this breaks the other functionality mentioned in the comments in this file.


lllyasviel:
after some consideration it seems that double print can show some info when users share logs
when user log have single print we will know something is wrong

Issue #1205, opened 2023-05-08
Feature Request - Image cropping in controlnet 1.1


jonwong666:
This was addressed in controlnet 1.0 , however with the new 1.1 release the cropping feature is gone.

https://github.com/Mikubill/sd-webui-controlnet/pull/483

Is it possible to reimplement cropping to 1.1 please?

Thank you

lllyasviel:
this can be really complicated considering we have inpaint mask now.
but note that if you do not give img to cn, cn can use img in i2i tab, and the img in i2i tab has a crop button

jonwong666:
> this can be really complicated considering we have inpaint mask now. but note that if you do not give img to cn, cn can use img in i2i tab, and the img in i2i tab has a crop button

thank you for the reply, I understand. However I use the old crop feature a lot in txt 2 img during my art making, to experiment with different sources /  references that may not be in the correct ratio or framing. This was a useful because I didn't have to preprocess the input and could drag and drop directly from browser which allowed for a more creative experience.

the solution implemented in the previous version by  @missionfloyd  used a separate tab for cropped images, which sent the result back to the main image window, this shouldn't mess with the inpaint?

I would really appreciate someone helping with this request. 

thank you

jonwong666:
anyone care to give it a shot?

jonwong666:
bump ( please )

PR #1334, opened 2023-05-19
Add inpaint upload mode for txt2img use

OedoSoldier:


lllyasviel:
can you share a screenshot about how it looks like?
and if the design looks good we will also consider adding gradio image crop in a similar way - like https://github.com/Mikubill/sd-webui-controlnet/issues/1205 https://github.com/Mikubill/sd-webui-controlnet/pull/483

lllyasviel:
unfortunately, mask blur does not make sence in CN inpaint  - controlnet inpaint must receive binarized mask
https://github.com/Mikubill/sd-webui-controlnet/blob/539d2fc990bc592dd7e482ecabffdc511073dbed/scripts/controlnet.py#L890

This can be merged after the mask blur slider is removed

OedoSoldier:
Oh, I overlooked that.

As for crop function, I think using `gr.State()` to store the original image can helps, maybe I'll try it later

lllyasviel:
after taking a look at this, crop is not quite related.

lllyasviel:
unfortunately we cannot modify ControlNetUnit for this - this change is too big and will affect all apis and other extensions.

lllyasviel:
can we remove all codes related to ControlNetUnit? they seems not related to the functionality and it can work without it

lllyasviel:
it seems that we need more considerations on this - although we want to avoid ControlNetUnit, if we do not do it using unit, then all units will share one input - if you use this mask upload in multiple units, bugs are likely to appear.
it seems that we do not have very good solution to this, and will consider this in next big update or try to find some solution

PR #1337, opened 2023-05-20
shift to context marker for cond-ucond

lllyasviel:
We will begin to use prompt marker to find out which instance is cond or ucond.
This will close https://github.com/pkuliyi2015/multidiffusion-upscaler-for-automatic1111/issues/153 https://github.com/pkuliyi2015/multidiffusion-upscaler-for-automatic1111/issues/191

### DIFF
--- a/scripts/controlnet.py
+++ b/scripts/controlnet.py
@@ -1200,8 +1200,6 @@ class Script(scripts.Script):
                 if unit.module == 'clip_vision':
                     detected_maps.append((processor.clip_vision_visualization(detected_map), unit.module))
 
-            is_vanilla_samplers = p.sampler_name in ["DDIM", "PLMS", "UniPC"]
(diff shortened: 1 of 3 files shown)


Lancer907:
Hello thanks you for your work.
I just updated to ControlNet v1.1.180 and I found a issue. When I open controlnet, the control mode and Resize Mode are default "none selected".
image
After I select it, if I do any change elsewhere, such as select Preprocessor or model or just Generate, the control mode and Resize Mode go back to "none selected ".

lllyasviel:
@Lancer907 update to 181, disable SD-CN-Animation, restart terminal,

Lancer907:
> @Lancer907 update to 181, disable SD-CN-Animation, restart terminal,

thanks! Problem solved

Issue #1238, opened 2023-05-13
[Feat]: cache from previous preprocessing


jjangga0214:
### Is there an existing issue for this?

- [X] I have searched the existing issues and checked the recent builds/commits of both this extension and the webui

### What happened?

Hi.
This is actually a feature request. I'm sorry if there's another proper place to suggest. I'll make sure to forward this there then.

The extension analyses a given image every time. 
It may be great if it caches the result by hash. 
So it can pass duplicate calculations when the same image is given again with the same setting.

How do you think? 

Thanks.

### Steps to reproduce the problem

.

### What should have happened?

.

### Commit where the problem happens

.

### Command Line Arguments

```Shell
.
```

### Console logs

```Shell
.
```

huchenlei:
#807 is intended to add caching for preprocessor results, but it did not made to merge, because of image hash issue.

If many people request this feature, I can try to submit a PR to implement it.

jjangga0214:
@huchenlei Thanks for the reply.
This is a definitely helpful feature, especially for cases where the same base image is used and used again hundreds of times.
I sincerely request this feature.
Thank you so much.

huchenlei:
Cache feature implemented in #1371.

PR #1371, opened 2023-05-21
[Enhancement] Cache preprocessor results

huchenlei:
#807 attempts to add a cache for preprocessor, but using an inefficient way:
- Use str(ndarray) as hash will make array with same start & end return same hash
- Each time the preprocessor is called, the caching decorator is reapplied

#1238 also requests us adding cache for preprocessor.

This PR adds an shared LRU cache of size 16 for all preprocessor functions. We can debate the cache size later. I set it to 16 as I feel like I often don't need longer history of preprocessor results.

Currently openpose preprocessor will not be cached when called with trigger button in preview, as each time the preprocessor receives a new instance of `JsonAcceptor`. We probably will want to rework the function signature of preprocessor later so that openpose preprocessor can be cached as well.

### DIFF
--- a/preload.py
+++ b/preload.py
@@ -1,4 +1,7 @@
 def preload(parser):
+    parser.add_argument("--controlnet-dir", type=str, help="Path to directory with ControlNet models", default=None)
+    parser.add_argument("--controlnet-annotator-models-path", type=str, help="Path to directory with annotator model directories", default=None)
+    parser.add_argument("--no-half-controlnet", action='store_true', help="do not switch the ControlNet models to 16-bit floats (only needed without --no-half)", default=None)
     # Setting default max_size=16 as each cache entry contains image as both key 
     # and value (Very costly).
     parser.add_argument(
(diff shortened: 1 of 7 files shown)


lllyasviel:
I am not familiar with this - it looks using many advanced python grammars
if merged you need to maintain this part and we will have an extra calling layer in calling stack everytime we debug preprocessors - so consider carefully before doing this

lllyasviel (approved):

lllyasviel:
but considering this can really speed up the cases where user only change prompts, I actually encourage to merge this - but water under bridge - test it a bit more since it affects all functionalities

lllyasviel:
and is it necessary to add a setting item to specify the cache counts? or since it is just numpy arrays, we just use 8 or 16?

huchenlei:
Thanks for the feedback!

Some changes I made to the PR:
- Cache size can now be specified in commandline argument when the user launches WebUI. I doubt anyone will need to adjust cache size on the fly.
- Setting cache size = 0 will completely remove the caching layer. This will be helpful during debug.
- API and UI now have separate cache so that the cache won't interfere with each other.

bienvanphuc:
hi @huchenlei
Is Openpose now supported to save cache in the latest version?

PR #1402, opened 2023-05-23
Fetch control_net_max_models_num settings from the api

AbdullahAlfaraj:
In the Auto-Photoshop plugin, we extract the number of ```control_net_max_models_num``` from the ```/config``` endpoint. However, this doesn’t always work due to user interface localization as in [this]( https://github.com/AbdullahAlfaraj/Auto-Photoshop-StableDiffusion-Plugin/issues/266). 
This pull request makes it easier for third-party tools to get the number of max units from the API.

### DIFF
--- a/scripts/api.py
+++ b/scripts/api.py
@@ -45,6 +45,11 @@ def controlnet_api(_: gr.Blocks, app: FastAPI):
             "module_list": _module_list,
             "module_detail": external_code.get_modules_detail(alias_names)
         }
+    
+    @app.get("/controlnet/settings")
+    async def settings():
+        max_models_num = external_code.get_max_models_num()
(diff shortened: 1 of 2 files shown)


ljleb:
This more a patch than a proper fix IMHO. We should not need this number to make API calls.

Instead of this, we should make a PR in the webui so that it accepts an arbitrary number of arguments for alwayson scripts.

As the webui development pace is not constant, I believe this is a good temporary fix for the limited functionality of the webui, but also I think a companion PR in the webui should be opened to fix the issue at the root.

lllyasviel:
This can be merged as demanded - but just want to mention that user interface localization happens at HTML level (it is HTML element text replacement, and the config files are not influenced)
if something wrong with config then the problem is in the reading logic

PR #1407, opened 2023-05-23
Fix m2m script to allow use of reference only

mattyamonaca:
Modified to accept not only video but also image input when using m2m script.


lllyasviel:
Can someone familiar with M2M review this?

mattyamonaca:
I don't think anyone else falls under this conditions since I'm the one who implemented M2M.
I have actually generated the video and confirmed that it works, but please let us know if there are any other checks I need to make.

Stan-Stani:
Does this mean it works with gifs?

PR #1416, opened 2023-05-24
feat: significantly accelerates detectmap_proc performance

storyicon:
I encountered performance issues with `shuffle` and `openpose`. After locating them, I found that in my tests:
1. openpose took 0.5s in detectmap_proc
2. shuffle takes 1.45s in detectmap_proc
and more than 95% of detectmap_proc's time consumption is concentrated in np.unique:
```
unique_color_count = np.unique(x.reshape(-1, x.shape[2]), axis=0).shape[0]
```
Now after optimization, `shuffle` takes only 0.11s on `detectmap_proc`, the time consumed by `openpose` on `detectmap_proc` is almost negligible.
I believe most processors can benefit from this optimization.

### DIFF
--- a/scripts/xyz_grid_support.py
+++ b/scripts/xyz_grid_support.py
@@ -5,7 +5,8 @@ from modules import scripts, shared
 
 try:
     from scripts.global_state import update_cn_models, cn_models_names, cn_preprocessor_modules
-    from scripts.external_code import ResizeMode
+    from scripts.external_code import ResizeMode, ControlMode
+
 except (ImportError, NameError):
     import_error = True
 else:
@@ -400,6 +401,9 @@ def add_axis_options(xyz_grid):
         update_cn_models()
         return list(cn_models_names.values())
 
+    def choices_control_mode():
+        return [e.value for e in ControlMode]
+
     def choices_resize_mode():
         return [e.value for e in ResizeMode]
 
@@ -413,6 +417,7 @@ def add_axis_options(xyz_grid):
 
     validation_data = [
         {"name": "model", "type": str, "check": choices_model, "exclude": make_excluded_list},
+        {"name": "control_mode", "type": str, "check": choices_control_mode, "exclude": None},
         {"name": "resize_mode", "type": str, "check": choices_resize_mode, "exclude": None},
         {"name": "preprocessor", "type": str, "check": choices_preprocessor, "exclude": None},
     ]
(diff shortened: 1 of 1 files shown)


storyicon:
The optimization would be more obvious if new dependencies, such as `cupy`, were allowed to be imported:
```
import cupy as cp

def cupy_count_unique_axis0(data):
    arr = cp.asarray(data)
    sorted_arr = arr[cp.lexsort(arr.T[::-1])]
    mask = cp.empty(arr.shape[0], dtype=cp.bool_)
    mask[0] = True
    mask[1:] = cp.any(sorted_arr[1:] != sorted_arr[:-1], axis=1)
    return len(sorted_arr[mask])
```
With this optimization, the `1.45s` time consumed by `np.unique` during `shuffle` can be optimized to `0.007s`.

I believe `cupy` can bring performance improvements to all aspects of controlnet, and if it is welcome, then I can submit optimizations for `cupy` on that basis.

huchenlei:
Thanks for spotting that formance issue!

The code LGTM. Just need a final stamp from @lllyasviel.

lllyasviel:
cool. will test it soon

lllyasviel:
So in numpy the function to get unique value counts in an array is O(N^2)

This method sort array and see how many elements are same as successive elements to count duplicated items, and then get unique value counts, so it is O(NlgN)

Looks good to me but is it possible to do it in O(N)?

huchenlei:
I think what lllyasviel means is use a hashset to track uniqueness instead of sorting. 

```
def get_unique_axis0_count(data):
    arr = np.asanyarray(data)
    idxs = np.lexsort(arr.T) <---
    arr = arr[idxs]
    unique_idxs = np.empty(len(arr), dtype=np.bool_)
    unique_idxs[:1] = True
    unique_idxs[1:] = np.any(arr[:-1, :] != arr[1:, :], axis=-1)
    return len(arr[unique_idxs])
```

Maybe something like:
```
def get_unique_axis0_count(data):
    return len(set(data))
```

storyicon:
There is an additional cost to constructing the hash table. In fact, using `set` is slightly better than `np.unique`, but much worse than `get_unique_axis0_count`.

storyicon:
@lllyasviel I noticed this PR has been going on for some time now, do you have any thoughts on moving forward with this optimization?

lllyasviel:
still considering this - it can speed up a bit when image is large (when image is small, I cannot feel any difference) but it also make codes more complicated and in fact we are not sure whether we will need other outputs from np.unique in the future. and even when image is large, the save time is < 1s and this func is only called once per-generating.

storyicon:
I use a 576x832 image on the Nvidia A10, and when I `shuffle` it, `np.unique` takes more than 1.4s.

PR #1422, opened 2023-05-25
[UI Update] The Control Type Filter is finally implemented

lllyasviel:
Beginning from 1.1.192, we will add a quick shortcut in UI, the Control Type filter.

image

You can turn off this quick filter in setting if you prefer not to have it, using the option “Disable control type selection”:

image

This filter has two functionalities:

First, if you select one type of control, your preprocessor list and model list will only show items of that type. You can always select “All” to see all preprocessors/models just like before. For example, I select “depth”

image

Then

image

Second, if you select one type of control, it will automatically select a default combination of that control. For example, I select “depth”

image

Then

image

Note that this filter has nothing to do with API and code logics so it will has zero influence to all other extensions.

Now we can swift methods much easier. And we do not need to worry about the list growing longer and longer.

**You need at least 1.1.192 to use it.**

### DIFF
--- a/scripts/controlnet.py
+++ b/scripts/controlnet.py
@@ -32,7 +32,7 @@ import base64
 from pathlib import Path
 from PIL import Image, ImageFilter, ImageOps
 from scripts.lvminthin import lvmin_thin, nake_nms
-from scripts.processor import preprocessor_sliders_config, flag_preprocessor_resolution, model_free_preprocessors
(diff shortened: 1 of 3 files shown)


huchenlei:
Can radio buttons be replaced with a dropdown?

If we expect more models coming in the future, the radio button design is not good because of scalability issue.

lllyasviel:
radio feels much better if u try both

PR #1449, opened 2023-05-27
Add classes/ids to some controls for styling

DrJKL:


scripts/controlnet.py
```diff
@@ -355,10 +355,10 @@ def refresh_all_models(*inputs):
         if shared.opts.data.get("controlnet_disable_control_type", False):
             type_filter = None
         else:
-            with gr.Row():
-                type_filter = gr.Radio(list(preprocessor_filters.keys()), label=f"Control Type", value='All', elem_id=f'{elem_id_tabname}_{tabname}_controlnet_type_filter_radio')
+            with gr.Row(elem_id="controlnet_control_type"):
```
huchenlei, line 358: I think use `elem_id` here will cause element id conflict among multiple controlnet ui-groups. Please replace `elem_id` with `elem_classes`. Same in line 361.

DrJKL: Okay, classes it is.

DrJKL:
For context, this sets up for this other PR: https://github.com/vladmandic/automatic/pull/1187

lllyasviel:
why we remove text "[Loopback]"? this may influence some language packages that match gradio texts

DrJKL:
The Loopback is still present, but as a label
image

lllyasviel:
yes I believe they match full string (at HTML level)

DrJKL:
Okay, reverted the label/info. I do think it'd be nicer to have a shorter name there. Are the localization files in their own repo?

lllyasviel:
ok merged

PR #1495, opened 2023-06-01
Add pre-commit check

huchenlei:
Request from #1485

This PR adds a pre-commit check for various things. It should help newly committed code to follow the same guideline.


lllyasviel:
Is this really necessary? Perhaps we get a bit more inputs from other devs @Mikubill @ljleb

JulianVolodia:
best is to do it once, and run all tests -- then just ask for redo some work of others (like - they could rebase just before the commit with format fix but with pre-commit setup, and then apply this precommit fixes to theirs work).

most of IDEs have those in place and people shouldnt lint their code

PR #1513, opened 2023-06-02
[Refactor] Move choose input image logic out of Script.process

huchenlei:
This PR moves the logic of picking the input image out of the gigantic `Script.process` function. 

Choose input image from following sources with descending priority:
 - p.image_control: ????
 - p.control_net_input_image: ????
 - unit.image: 
   - ControlNet tab input image.
   - Input image from API call.
 - p.init_images: A1111 img2img tab input image.

Can someone explain what are the first 2 branches for?

### DIFF
--- a/scripts/controlnet.py
+++ b/scripts/controlnet.py
@@ -498,8 +498,8 @@ class Script(scripts.Script):
             idx: int
         ) -> Tuple[np.ndarray, Optional[external_code.ResizeMode]]:
         """ Choose input image from following sources with descending priority:
-         - p.image_control: ????
-         - p.control_net_input_image: ????
+         - p.image_control: [Deprecated] Lagacy way to pass image to controlnet.
+         - p.control_net_input_image: [Deprecated] Lagacy way to pass image to controlnet.
          - unit.image: 
            - ControlNet tab input image.
            - Input image from API call.
@@ -515,8 +515,10 @@ class Script(scripts.Script):
         image = image_dict_from_any(unit.image)
 
         if batch_hijack.instance.is_batch and getattr(p, "image_control", None) is not None:
+            print("Warn: Using legacy field 'p.image_control'.")
             input_image = HWC3(np.asarray(p.image_control))
         elif p_input_image is not None:
+            print("Warn: Using legacy field 'p.controlnet_input_image'")
             if isinstance(p_input_image, dict) and "mask" in p_input_image and "image" in p_input_image:
                 color = HWC3(np.asarray(p_input_image['image']))
                 alpha = np.asarray(p_input_image['mask'])[..., None]
--- a/scripts/controlnet.py
+++ b/scripts/controlnet.py
@@ -2,7 +2,7 @@ import gc
 import os
 from collections import OrderedDict
 from copy import copy
-from typing import Dict, Optional
+from typing import Dict, Optional, Tuple
 import importlib
 import modules.scripts as scripts
 from modules import shared, devices, script_callbacks, processing, masking, images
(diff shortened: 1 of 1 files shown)


ljleb:
Setting properties on `p` is the old way of passing values to controlnet from other scripts. We may want to eventually drop support for this to simplify the code. However some scripts may still use it, so I am not sure.

Maybe should make an announcement and if nobody complains we can try to drop support?

Reducing the size of the process method is a good idea IMO. It makes it easier to understand what is going on and eventually we will be able to test the code.

huchenlei:
> Setting properties on `p` is the old way of passing values to controlnet from other scripts. We may want to eventually drop support for this to simplify the code. However some scripts may still use it, so I am not sure.
> 
> Maybe should make an announcement and if nobody complains we can try to drop support?
> 
> Reducing the size of the process method is a good idea IMO. It makes it easier to understand what is going on and eventually we will be able to test the code.

Is the first branch also deprecated? 
`batch_hijack.instance.is_batch and getattr(p, "image_control", None)` seems to be batch process specific.

ljleb:
It seems to be batch specific, however it was introduced before the batch update: https://github.com/Mikubill/sd-webui-controlnet/commit/70d4e34ecd0cb691acff529ae09af262f5f02dce

ljleb:
Looking at entire github, we can see that not many github repos use the `p.image_control`: https://github.com/search?q=p.image_control&type=code&p=1

However, the story seems to be different for `p.control_net_input_image`: https://github.com/search?q=p.control_net_input_image&type=code

That's a lot of code. It seems wiser to keep the latter for the time being.

continue-revolution:
My extension https://github.com/continue-revolution/sd-webui-segment-anything also pass value to cnet via control_net_input_image. I honestly don’t think this is a good way since it depends on the script priority inside A1111. I want to know the way you recommend.

ljleb:
To my understanding, the external code API should work for this use case. Did you run into any issue using it?

continue-revolution:
Seems that still only scripts that have a "smaller" folder name can control controlnet inside `process` and there is no good solution (at least in my mind) to overcome this problem. My script bypassed this problem in a very ugly way ([here](https://github.com/continue-revolution/sd-webui-segment-anything/blob/master/scripts/sam.py#L538)->[here](https://github.com/continue-revolution/sd-webui-segment-anything/blob/master/scripts/sam.py#L407-L426)). I tend to stick to the old way of passing value to cnet via `control_net_input_image` before a better solution is discovered.

ljleb:
To my understanding, swapping the order of cnet with your script is orthogonal to using `p.control_net_input_image` vs extrnal code api. As the api finds cnet by name, it will still find cnet and set the correct args in p.script_args.

Did you run into issues using the swap trick + external code api?

continue-revolution:
yes. whether or not using the swap trick is unrelated to whether or not using the external code api. I have not used your external code api, but the code looks good to me, so I don't believe it will cause any trouble.

ljleb:
The reason the code api was created was so that we could manage internal implementation details without breaking other extensions. By calling a function that is defined in this repo instead of modifying a property on `p`, it is much easier to change the implementation whenever we need to.

I agree replacing code that does one thing with other code that does exactly the same thing looks pointless. For now you won't get much benefits. The reason I believe people should move to the supported api anyways, is that if we end up needing to do something that cannot be done by setting properties on `p`, then there won't be much you will be able to do if we update controlnet code. I don't have an example now, but code changes all the time so it is a non-zero possibility.

In any case, whether this convinces you or not, to be able to remove this dependency on our end, I would need to convince a good number of other people that also use the properties on `p` to send values to controlnet. I'm not sure it is practical to go there, and for the foreseeable future there doesn't seem to be things we need to do that we can do with a function call but cannot with the properties on `p`.

PR #1522, opened 2023-06-02
:hammer: Extract pixel perfect calculation

huchenlei:
This PR extracts the duplicated pixel perfect calculation out as a function. 

Note: docstring generated by ChatGPT (GPT4).


lllyasviel:
print(f"Pixel Perfect Mode Enabled In Preview.") vs print(f"Pixel Perfect Mode Enabled.") ?
Perhaps we always 
print(f"Pixel Perfect Computation:")

PR #1523, opened 2023-06-02
Add logger

huchenlei:
This PR adds a logger to help distinguish different levels of logs. The default log level is set to `INFO`. 

Example log of ControlNet version:
```
2023-06-02 18:01:51,483 - ControlNet - INFO - ControlNet v1.1.213
```

We should be able to slowly migrate to using the logger from unconditional print.

### DIFF
--- a/scripts/logging.py
+++ b/scripts/logging.py
@@ -1,0 +1,23 @@
+import logging
+
+from modules import shared
+
+# Create a new logger
+logger = logging.getLogger("ControlNet")
+
+# Configure logger
+handler = logging.StreamHandler()
+handler.setFormatter(
+    logging.Formatter("%(asctime)s - %(name)s - %(levelname)s - %(message)s")
+)
+logger.addHandler(handler)
+
+
+if hasattr(shared.cmd_opts, "controlnet_loglevel"):
+    loglevel_string = shared.cmd_opts.controlnet_loglevel
+else:
+    print("Warning: `controlnet_loglevel` not available from A1111")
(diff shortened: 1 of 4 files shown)


scripts/logging.py
```diff
@@ -0,0 +1,23 @@
...
+    logging.Formatter("%(asctime)s - %(name)s - %(levelname)s - %(message)s")
+)
+logger.addHandler(handler)
+
+
+if hasattr(shared.cmd_opts, "controlnet_loglevel"):
+    loglevel_string = shared.cmd_opts.controlnet_loglevel
+else:
+    print("Warning: `controlnet_loglevel` not available from A1111")
+    loglevel_string = "DEBUG"
```
ljleb, line 20: Why not only use `getattr(shared.cmd_opts, "controlnet_loglevel", "INFO")`? AFAIK the if branch should only run when users do not do a full restart of the webui. Also, is there a reason we are falling back on debug here but using info by default?

ljleb:
Looks good overall, although we should probably reload scripts/logging.py in scripts/controlnet.py
I have seen some efforts on the side of a1111 webui to use logging as well. However I am not sure where they are at with this and whether a part of the cake will be usable by extensions.

huchenlei:
I think we should be fine as we are not directly using the root logger. I'd be happy if A1111 adds a global loglevel cmd argument that we can inherit from.

scripts/logging.py
huchenlei: Done.

PR #1525, opened 2023-06-03
Extract prepare mask logic

huchenlei:


lllyasviel:
This pull breaks "inpaint only masked"+leave CN input blank.

image
image
image

Loading model from cache: control_v11p_sd15_inpaint [ebff9138]
Error running process: D:\sd-webui-aki-v4\extensions\sd-webui-controlnet\scripts\controlnet.py
  File "D:\sd-webui-aki-v4\modules\scripts.py", line 451, in process
  File "D:\sd-webui-aki-v4\extensions\sd-webui-controlnet\scripts\controlnet.py", line 697, in process
    mask = prepare_mask(a1111_mask, p)
  File "D:\sd-webui-aki-v4\extensions\sd-webui-controlnet\scripts\controlnet.py", line 159, in prepare_mask
    mask = mask.convert("L")
AttributeError: 'numpy.ndarray' object has no attribute 'convert'

lllyasviel:
reverted as emergency fix

huchenlei:
Seems like `a1111_mask` is reused, and the type changed.

lllyasviel:
consider revise and use revised version

lllyasviel:
should be fixed if prepare_mask only receive p and read mask inside function

huchenlei:
I am going to rename the Image type of `a1111_mask` to `a1111_mask_image`, so that type of variable does not change through out the context.

lllyasviel:
I think as long as the above test case works, whatever is ok. I only know this can influence the "inpaint only masked"
Also, "ultimate SD upscaler" is hacking this part of codes to support CN, which is an inpaint-based hack, perhaps also test that if this part is going to change

PR #1539, opened 2023-06-04
remove unnecessary/wrong judgement

lllyasviel:


huchenlei:
Can you take a look at some other conditions in `process`?

I find some conditions really hard to reason:

```
if 'inpaint' in unit.module and not image_has_mask(input_image) and a1111_mask_image is not None:
    ...
```

```
if 'reference' not in unit.module and issubclass(type(p), StableDiffusionProcessingImg2Img) \
                    and p.inpaint_full_res and a1111_mask_image is not None:
    ...
```

They seem to apply a1111 mask in different ways, but not strictly excluding eachother.

Also the `resize_mode` value is vague. So only when we use image from a1111 or mask from a1111 that we use a1111's resize mode?

Issue #1144, opened 2023-05-03
[Feature Request]: Add pixel_perfect to /controlnet/detect


Hetaneko:
I'm using ControlNet v1.1.133.
The current payload is this:
![image](https://user-images.githubusercontent.com/54069666/235956611-ae7117fc-1015-49cb-90b6-9e2707a58349.png)

Since we already have pixel perfect mode on /controlnet/*2img. Can we have it as well as on the /controlnet/detect

kft334:
I did this or something similar in my fork using a new endpoint and also added transparency support but I didn't make a PR for it. The fork isn't up-to-date but feel free to rip the added endpoint.
 
https://github.com/kft334/sd-webui-controlnet/commit/cab797404a003002e638da1a9a8d26701dac021b

lllyasviel:
in api, we do not have access to the width/height of gradio slider of you need to set resolution with script on your own. it is just the short side of your image

PR #1545, opened 2023-06-04
Override detection res if pixel perfect is set

ramyma:
Added `pixel_perfect` (defaults to `False`) to the `detect` endpoint which accepts `resize_mode` as well (defaults to `Just Resize`).
The change is backwards compatible.

I tried updating the tests but couldn't get them running locally.
Would be nice to have a guide on how to run tests.


huchenlei:
I will also appreciate if someone can explain how to run the api tests as well. 
```
python launch.py --no-half --disable-opt-split-attention --use-cpu all --skip-torch-cuda-test --api --tests ./extensions/sd-webui-controlnet/tests
```
Running the command locally says there is no such commandline option `--tests`.

ljleb:
It seems the webui tests have been updated 2 weeks ago: https://github.com/AUTOMATIC1111/stable-diffusion-webui/blob/master/.github/workflows/run_tests.yaml
Currently the controlnet tests workflow uses an old version of the webui to run tests against: 5ab7f213bec2f816f9c5644becb32eb72c8ffb89
We will eventually need to adapt their workflow so that it works here

scripts/api.py
```diff
@@ -77,8 +79,29 @@ async def detect(
...
             img = external_code.to_base64_nparray(input_image)
-            results.append(processor_module(img, res=controlnet_processor_res, thr_a=controlnet_threshold_a, thr_b=controlnet_threshold_b)[0])
+
+            res = controlnet_processor_res
+
+            if controlnet_pixel_perfect:
+
+                h, w = img.shape[:2]
+
+                h = (h // 8) * 8
```
huchenlei, line 91: The casting is not pixel perfect specific in `process`. 

I believe this casting does not appear in `run_annotator` code in uigroup. 
@lllyasviel Is this logic necessary in `run_annotaor` as well?

ramyma: BTW, I borrowed this logic from existing code https://github.com/Mikubill/sd-webui-controlnet/blob/09cb9a32d1051aa827f1bb092cf17fcbf996ed7f/scripts/controlnet.py#L1204

ramyma: I added a couple of tests as well.

ramyma:
I added a couple of tests.

lllyasviel:
I think this computation is not very correct though - the pixel-perfect is targeted to the diffusion resolution, not the image resolution. For example, if the diffusion resolution is 512x768 and the input image is 800x800, then if "crop and resize", the estimation is 768 (resize to 768x768 and then crop to 512x768). if "resize and fill", the estimation is 512 (resize to 512x512 and then pad/fill to 512x768).
This computation is not possible without knowing diffusion parameters

scripts/api.py
huchenlei: > BTW, I borrowed this logic from existing code
> 
> https://github.com/Mikubill/sd-webui-controlnet/blob/09cb9a32d1051aa827f1bb092cf17fcbf996ed7f/scripts/controlnet.py#L1204

The existing code is outside `if unit.pixel_perfect:` check.

ramyma:
> I think this computation is not very correct though - the pixel-perfect is targeted to the diffusion resolution, not the image resolution. For example, if the diffusion resolution is 512x768 and the input image is 800x800, then if "crop and resize", the estimation is 768 (resize to 768x768 and then crop to 512x768). if "resize and fill", the estimation is 512 (resize to 512x512 and then pad/fill to 512x768). This computation is not possible without knowing diffusion parameters

Well, I believe that in the context of an endpoint call only the image being passed is relevant.
The way I see/treat the detect endpoint is as a gateway for getting images through preprocessors, rather than being concerned with the whole pipeline, unlike the UI implementation.

ramyma:
@lllyasviel what are your thoughts?

ramyma:
Anyone?

huchenlei:
> Anyone?

I guess you don't get lllyasviel's point. Pixel Perfect is only relevant in the context of Width and Height of final generated image, while the `detect` API has no such context. Using the controlnet input's Width and Height is not the correct way of Pixel Perfect.

ramyma:
> > Anyone?
> 
> I guess you don't get lllyasviel's point. Pixel Perfect is only relevant in the context of Width and Height of final generated image, while the `detect` API has no such context. Using the controlnet input's Width and Height is not the correct way of Pixel Perfect.

What is your suggestion to get an endpoint detection using the input image dimensions?
Passing `res` explicitly is not very practical, I just thought reusing the existing logic would be close enough. 
In my tests it gave me reasonable results for this use case.
Perhaps exposing `resize_method` is not necessary in this case; I can hard code it to just resize.

huchenlei:
> 

I think if you really want pixel perfect to be part of the `detect` API, you need to pass in the Width and Height intead of using the controlnet input's width and height.

ramyma:
I updated the PR to use the dimensions of `input_image` to override `res` when `pixel_perfect` is `True`, removed `resize_mode` from the equation.

ramyma:
@lllyasviel @huchenlei can you please review?

huchenlei:
See https://github.com/Mikubill/sd-webui-controlnet/blob/4f884009efddf72c33493e3a7927f04817a719d7/scripts/controlnet_ui/controlnet_ui_group.py#L674.

The implementation in `controlnet_ui_group` get the width and height from A1111 sliders, which control the dimensions of the final output image.

huchenlei:
Issue already discussed in
#1144

PR #1550, opened 2023-06-04
Added support for control mode to xyz grid

Ezekiel-Rage:
Added support for control mode to xyz grid


Ezekiel-Rage:
Screenshot in case the comment above isn't clear. Control mode was inexplicably missing in xyz grid. This PR adds it.

control_mode

huchenlei:
LGTM. This is indeed one thing useful that is missing right now.

Issue #1541, opened 2023-06-04
txt2img tests silently broke


huchenlei:
See this tryjob:
https://github.com/Mikubill/sd-webui-controlnet/actions/runs/5165879820/jobs/9305660056

I changed `print` to `logger.error` and the build fails. 

Previously we were only checking `'error'` in `stderr`, while we were outputing our own error message to `stdout`. 

```
Traceback (most recent call last):
  File "/home/runner/work/sd-webui-controlnet/sd-webui-controlnet/stable-diffusion-webui/extensions/sd-webui-controlnet/tests/web_api/txt2img_test.py", line 130, in test_call_with_preprocessors
    self.assert_status_ok(f'Running preprocessor module: {module}')
  File "/home/runner/work/sd-webui-controlnet/sd-webui-controlnet/stable-diffusion-webui/extensions/sd-webui-controlnet/tests/web_api/txt2img_test.py", line 77, in assert_status_ok
    self.assertFalse('error' in stderr, "Errors in stderr: \n" + stderr)

2023-06-03 22:22:44,537 - controlnet - info - loading model: control_canny-fp16 [e3fe7712]
2023-06-03 22:22:44,591 - controlnet - info - loaded state_dict from [/home/runner/work/sd-webui-controlnet/sd-webui-controlnet/stable-diffusion-webui/extensions/sd-webui-controlnet/models/control_canny-fp16.safetensors]
2023-06-03 22:22:44,593 - controlnet - error - error: controlnet cannot find model config [/home/runner/work/sd-webui-controlnet/sd-webui-controlnet/stable-diffusion-webui/extensions/sd-webui-controlnet/models/control_canny-fp16.yaml] 
error: controlnet will use a wrong config [/home/runner/work/sd-webui-controlnet/sd-webui-controlnet/stable-diffusion-webui/extensions/sd-webui-controlnet/models/cldm_v15.yaml] to load your model. 
error: the wrong config may not match your model. the generated results can be bad. 
error: you are using a controlnet model [control_canny-fp16] without correct yaml config file. 
error: the performance of this model may be worse than your expectation. 
error: if this model cannot get good results, the reason is that you do not have a yaml file for the model. 
solution: please download yaml file, or ask your model provider to provide [/home/runner/work/sd-webui-controlnet/sd-webui-controlnet/stable-diffusion-webui/extensions/sd-webui-controlnet/models/control_canny-fp16.yaml] for you to download.
[...]

PR #1567, opened 2023-06-06
feat: significantly accelerates detectmap_proc performance

storyicon:
I found that with the change to the master branch, there were some conflicts with the previous commit https://github.com/Mikubill/sd-webui-controlnet/pull/1416. I made some adjustments and recommitted this optimization.

### DIFF
--- a/scripts/controlnet.py
+++ b/scripts/controlnet.py
@@ -19,7 +19,7 @@ importlib.reload(batch_hijack)
 from scripts.cldm import PlugableControlModel
 from scripts.processor import *
 from scripts.adapter import PlugableAdapter
-from scripts.utils import load_state_dict
+from scripts.utils import load_state_dict, get_unique_axis0
 from scripts.hook import ControlParams, UnetHook, ControlModelType
 from scripts.controlnet_ui.controlnet_ui_group import ControlNetUiGroup, UiControlNetUnit
 from scripts.logging import logger
@@ -431,7 +431,7 @@ class Script(scripts.Script):
 
             new_size_is_smaller = (size[0] * size[1]) < (x.shape[0] * x.shape[1])
             new_size_is_bigger = (size[0] * size[1]) > (x.shape[0] * x.shape[1])
-            unique_color_count = np.unique(x.reshape(-1, x.shape[2]), axis=0).shape[0]
+            unique_color_count = len(get_unique_axis0(x.reshape(-1, x.shape[2])))
             is_one_pixel_edge = False
             is_binary = False
             if unique_color_count == 2:
--- a/scripts/utils.py
+++ b/scripts/utils.py
@@ -108,3 +108,11 @@ def svg_preprocess(inputs: Dict, preprocess: Callable):
         inputs["image"] = base64_str
     return preprocess(inputs)
 
+def get_unique_axis0(data):
+    arr = np.asanyarray(data)
+    idxs = np.lexsort(arr.T)
+    arr = arr[idxs]
+    unique_idxs = np.empty(len(arr), dtype=np.bool_)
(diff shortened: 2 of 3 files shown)


storyicon:
I found in my testing that the optimization it brings is in some cases more obvious than turning on xformers, close to about 1s of optimization can save some companies a lot of graphics card costs, I think it is an effective optimization. @lllyasviel

storyicon:
cc:  @Mikubill

JulianVolodia:
I testing this for more than week. It works well for me. I merged into my fork your work. Thanks @storyicon for this! :)

huchenlei:
From https://github.com/Mikubill/sd-webui-controlnet/issues/1648 we know that `detect_map` can take long when processing large images. 2s for 3k x 2k image. 

@lllyasviel We should consider merging this to alleviate the issue.

lllyasviel:
I will take a look at this and hopefully this weekend. Hopefully it will not cause strange problems on mac and ndarray contiguous

lllyasviel:
Also how reliable is our unit test system now? Does passing unit test really means the test are passed with all assets succeed? cuz I remember a period of time someone says some errors will be hidden by the github ci

ljleb:
One thing that makes tests unreliable is that they are run on an old version of the webui.

AFAIK you are referring to web api tests that resulted in hidden failures, which should have been resolved by testing for the presence of the string "ERROR" in the logs.

huchenlei:
> One thing that makes tests unreliable is that they are run on an old version of the webui.
> 
> AFAIK you are referring to web api tests that resulted in hidden failures, which should have been resolved by testing for the presence of the string "error" in the logs.

I think we are only checking string "error" in stderr. The log messages ControlNet emitts are all on stdout. I think following error is triggered in our API tests but ignored:
```
        if os.path.exists(override_config):
            network_config = override_config
        else:
            # Note: This error is triggered in unittest, but not caught.
            # TODO: Replace `print` with `logger.error`.
            print(f'ERROR: ControlNet cannot find model config [{override_config}] \n'
                  f'ERROR: ControlNet will use a WRONG config [{network_config}] to load your model. \n'
                  f'ERROR: The WRONG config may not match your model. The generated results can be bad. \n'
                  f'ERROR: You are using a ControlNet model [{model_stem}] without correct YAML config file. \n'
                  f'ERROR: The performance of this model may be worse than your expectation. \n'
                  f'ERROR: If this model cannot get good results, the reason is that you do not have a YAML file for the model. \n'
                  f'Solution: Please download YAML file, or ask your model provider to provide [{override_config}] for you to download.\n'
                  f'Hint: You can take a look at [{os.path.join(global_state.script_dir, "models")}] to find many existing YAML files.\n')
```

ljleb:
> I think we are only checking string "error" in stderr.

That is true, and it could be a problem for errors that are printed to stdout. Maybe add file=sys.stderr for these prints?

I am not sure if testing for "error" in stdout could result in false failures? Maybe not. If not, verifying the content of stdout would work I think.

huchenlei:
For reference: #1541 has detailed error message

ljleb:
The truth is, reading console output to determine test failure is a hack. There should be a better way to do this. If it's out of the control of extensions, an issue needs to be raised on a1111 webui.

IIUC a proper fix could be to allow to disable extension error catching from a1111 webui code with a setting. I think I'll open an issue for that.

storyicon:
> I testing this for more than week. It works well for me. I merged into my fork your work. Thanks @storyicon for this! :)

I'm glad to hear that and happy to be able to contribute to the controlnet community.

huchenlei:
The PR passes web_tests locally. Going to merge it. If it breaks anything platform specific, feel free to revert it.

PR #1577, opened 2023-06-07
:art: Make active unit's tab name green

huchenlei:
When the user has a lot of controlnet units enabled. It can be hard to figure out what unit is enabled, what unit is not. Users have to check each unit tab and see if `enable` is checked. 

This PR adds the feature that makes active unit's tab name green so that users can easily know which unit is enabled. 

Effect with light theme:
Screen Capture 006 - Stable Diffusion - 127 0 0 1

Effect with dark theme:
Screen Capture 007 - Stable Diffusion - 127 0 0 1

Gradio recreate tab nav buttons every time active tab switches, which makes this feature implementation very hacky, but I am satisfied with the result.


huchenlei:
Now use brighter green in dark mode to assist reading:
Screen Capture 008 - Stable Diffusion - 127 0 0 1

PR #1603, opened 2023-06-09
fix: subscribe gr.Slider's change event, not release event

dennissheng:
If only subscribe gr.Slider's release event, when change value directly, no event will be triggered. But if  subscribe gr.Slider's change event, both release even and  change event will be triggerd.

Test code below:

```
import gradio as gr

with gr.Blocks() as demo:
    s = gr.Slider(2, 20, value=4, label="Count", info="Choose betwen 2 and 20", interactive=True)
    s.change(lambda : print("change"))
    # s.release(lambda: print("release"))
demo.launch()
```

### DIFF
--- a/scripts/controlnet_ui/controlnet_ui_group.py
+++ b/scripts/controlnet_ui/controlnet_ui_group.py
@@ -834,8 +834,6 @@ class ControlNetUiGroup(object):
                 event_subscribers.append(comp.edit)
             elif hasattr(comp, "click"):
                 event_subscribers.append(comp.click)
-            elif isinstance(comp, gr.Slider) and hasattr(comp, "release"):
-                event_subscribers.append(comp.release)
             elif hasattr(comp, "change"):
                 event_subscribers.append(comp.change)
 


lllyasviel:
what is changing value directly? change value with typing or do you mean change value using codes?

lllyasviel:
it seems that if I change with typing, release will be triggered when loss focus

dennissheng:
> what is changing value directly? change value with typing or do you mean change value using codes?

change value with typing

dennissheng:
> Semi-related to [AUTOMATIC1111/stable-diffusion-webui#10735](https://github.com/AUTOMATIC1111/stable-diffusion-webui/pull/10735). I PR'd this in #1316 because it improves performance (and in the webui in [AUTOMATIC1111/stable-diffusion-webui#10519](https://github.com/AUTOMATIC1111/stable-diffusion-webui/pull/10519)).
> 
> This certainly should ideally not break anything but it's also unfortunate to lose performance if we have to revert this.

But without listening to change event, when we type value and then click generate button, we will get the previous value stored in gr.State, which really make me confused. Maybe we can figure out a way to perfectly solve this problem.

dennissheng:
> it seems that if I change with typing, release will be triggered when loss focus

I'm not quite sure what you mean. In my environment (gradio 3.23.0), typing and mouse clicking of the bar both trigger change event, so the value stored in gr.State can be updated.

dennissheng:
> > gradio 3.23.0
> 
> [The release event was fixed in 3.23.1b1](https://github.com/gradio-app/gradio/pull/3589). Maybe that's what's causing the behavior you're seeing? The webui requirement is actually way ahead of this now and expects you're running 3.32.0. https://github.com/AUTOMATIC1111/stable-diffusion-webui/blob/baf6946e06249c5af9851c60171692c44ef633e0/requirements_versions.txt#L6

Thanks, you are right. I updated gradio, and then problem solved.

PR #1667, opened 2023-06-22
[API]: expose the filter functionality to api

AbdullahAlfaraj:
This PR refactors the filter preprocessor functionality  to work with both Gradio and the API


huchenlei:
Can you add a unittest for the newly added API?

AbdullahAlfaraj:
> Can you add a unittest for the newly added API?

I've added a basic unit test for when the user provides invalid keyword.

huchenlei:
I think the control types are pretty stable. There is probably no need to frequently fetch through API. Get all control types in one go should be better. I have changed the code do return all control type details.

PR #1695, opened 2023-06-23
:bug: Allow functionally equiv stale units

huchenlei:


lllyasviel:
I dont think we should give stale warnings when the unit is hacked by 3rd extensions?

huchenlei:
> I dont think we should give stale warnings when the unit is hacked by 3rd extensions?

Maybe demote that log to debug? I think it could be useful during debug, but not so useful for a common user.

lllyasviel:
and it should be safer to just make sure all keys in the toy unit exist in the instance unit, and ignore extra keys, as 3rd extensions would add extra keys to the unit

lllyasviel:
and probably ignore any key begin with "__" to avoid strange problems

lllyasviel:
like some one hacked the class and write a "__doc__" in the class then it will take one year to find out where is wrong and why controlnet cannot read the unit

huchenlei:
> 

`vars` do not return keys begin with "__", `dir` does that.

lllyasviel:
oh that make sense

PR #1707, opened 2023-06-25
Enable pixel perfect by default

huchenlei:
This PR enables pixel perfect mode by default for all ControlNet units. 

As far as I know in most situations I need to check pixel perfect checkbox. If there is no downside of using pixel perfect mode, we should make it enabled by default.

Correct me if there is any side effect/downside which makes pixel perfect mode an undesirable option.

### DIFF
--- a/scripts/controlnet.py
+++ b/scripts/controlnet.py
@@ -818,6 +818,11 @@ class Script(scripts.Script):
                     resize_mode=resize_mode
                 )
 
+            midas_modules = ("depth", "normal_map")
+            if unit.module in midas_modules and preprocessor_resolution > 512:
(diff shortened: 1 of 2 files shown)


lllyasviel:
midas's performance is bad when bigger than 512
it is trained at 384

huchenlei:
I add a log message reminding the user about midas's resolution. Am I understanding the context correctly? @lllyasviel

lllyasviel:
i am not sure. different preprocessors have different pretrained resolutions. using default value is a robust choice that will not cause problem, but sometimes suboptimal. using pixel perfect by default is overly aggressive

PR #1730, opened 2023-06-29
Update controlnet.py

Tomlirx:
temp fix a bug which block my use of reference ControlNet


huchenlei:
Can you share a step of reproduction of this issue? I don't think we are merging this PR if the context is not clear.

PR #1747, opened 2023-07-03
:sparkles: Display A1111 input as fallback image in img2img

huchenlei:
I found that many people don't know that if you do not upload an image in the ControlNet UI in img2img mode, the img2img input will be used as input to ControlNet.

This PR address this issue by setting the A1111 img2img input as background image in all ControlNet units. This will allow user realize that A1111 input will be used as fallback.

[screen-capture (3).webm](https://github.com/Mikubill/sd-webui-controlnet/assets/20929282/08d136ad-b31d-46c3-9557-5495ae00bbf2)

Note:
One drawback is that the user might click `run preprocessor` button with A1111 input as background and nothing happens. I think one way to address this is to display an error message on UI saying that no ControlNet input image.

### DIFF
--- a/javascript/img2img_input_display.js
+++ b/javascript/img2img_input_display.js
@@ -1,0 +1,157 @@
+/**
+ * Display the currently active img2img tab's input image as an uninteractable
+ * background image behind ControlNet's image input. This change will hint user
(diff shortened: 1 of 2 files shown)


huchenlei:
Note: web tests verified that all basic functionality holds after the change.

PR #1772, opened 2023-07-07
Update batch_hijack.py

apsntian:
change the comparison method of InputMode 

I'm modifying the webui code to use controlnet+adetailer as rest api, I found that this code causes problem. 

The problem is the class variable of InputMode is instantiated for each thread (UI and Rest API), so the comparison from Rest API doesn't work properly. In Python, it is not regarded as string, so the comparison is done with their id.  so I changed the comparison code into just simple string comparison.

### DIFF
--- a/scripts/batch_hijack.py
+++ b/scripts/batch_hijack.py
@@ -187,7 +187,7 @@ def get_cn_batches(p: processing.StableDiffusionProcessing) -> Tuple[bool, List[
     output_dir = ''
     input_file_names = []
     for unit in units:
-        if getattr(unit, 'input_mode', InputMode.SIMPLE) == InputMode.BATCH:
+        if str(getattr(unit, 'input_mode', InputMode.SIMPLE)) == str(InputMode.BATCH):
             any_unit_is_batch = True
             output_dir = getattr(unit, 'output_dir', '')
             if isinstance(unit.batch_images, str):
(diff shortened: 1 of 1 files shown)


huchenlei:
I don’t think it is because different thread will reinitialize the variable. This is a module reload problem similar to ControlNetUnit issue we had earlier. 

The comparison fail is also not because of ID comparison. It’s because the new class of input mode is different from the old input mode class.

If input mode is constructed within ControlNet’s scope, we should find a place to do proper module reload instead of comparing string values.

Note: a1111 has claimed to do the proper module reloading now after 1.4.0. Should we get rid of all module reload code so they don’t cause us troubles?

@ljleb @lllyasviel

ljleb:
These double definitions are a bit annoying.

> Note: a1111 has claimed to do the proper module reloading now after 1.4.0. Should we get rid of all module reload code so they don’t cause us troubles?

I think this is fine. The reload button in the extensions tab was also replaced by "apply and quit".

As we don't need to reload the modules now, a fix is to move all definitions of external_code.py to another file, i.e. internal_controlnet/external_code.py and then import * from that file. This should ensure that module caching from python will work properly.

ljleb:
@apsntian can you verify that #1775 fixes the issue? It has been merged in main just now.

apsntian:
> @apsntian can you verify that #1775 fixes the issue? It has been merged in main just now.

I checked that #1775 fixes my issue.  Thanks a lot!

PR #1774, opened 2023-07-07
Add preprocessor for PixelNet.

thomaseding:
Adds preprocessor support for PixelNet model.
https://huggingface.co/thomaseding/pixelnet
https://civitai.com/models/102482/pixelnet

### DIFF
--- a/annotator/pixelnet/__init__.py
+++ b/annotator/pixelnet/__init__.py
@@ -1,0 +1,49 @@
+import cv2
+import enum
+import numpy as np
+
+from PIL import Image
+from typing import Optional
+
+class Color(enum.Enum):
+    BLACK = (0, 0, 0)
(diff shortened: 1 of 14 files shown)


thomaseding:
This PR is ready. I'm just force-pushing it to keep on top of your latest commits.

thomaseding:
Some notes:

- The PixelNet preprocessor does not process input images. If any are supplied, they are completely ignored.

- "Pixel Perfect" ignores the Preprocessor's "Image Width" and "Image Height" and instead uses the dimensions provided by text-to-image.

- If you are not using "Pixel Perfect" and the dimensions don't match, you may find "Crop and Resize" to be more intuitive than "Just Resize". You can observe the preview image to see when and how they differ.

- Like other the other preprocessors, this serializes the PixelNet-specific slider information to the image metadata.

PR #1775, opened 2023-07-08
Remove `importlib.reload`s

ljleb:
Note: this only works with webui>=1.4.0 and IIUC on recent versions of vlad's fork.

Remove `importlib.reload`s and move the contents of external_code.py to another location so that python caching works, even when it is imported from another extension using `importlib.import_module`


scripts/external_code.py
```diff
@@ -1,453 +1 @@
...
-            return script
-
-
-def is_cn_script(script: scripts.Script) -> bool:
-    """
-    Determine whether `script` is a ControlNet script.
-    """
-
-    return script.title().lower() == 'controlnet'
+from internal_controlnet.external_code import *
```
huchenlei, line 1: Do we keep `scripts/external_code.py` here for compatibility. Do other extensions import this module directly?

tests/external_code_api/importlib_reload_test.py
```diff
@@ -0,0 +1,24 @@
...
+        importlib.reload(external_code)
+        NewControlNetUnit = external_code.ControlNetUnit
+        self.assertEqual(self.ControlNetUnit, NewControlNetUnit)
+
+    def test_force_import_does_not_redefine(self):
+        external_code_copy = importlib.import_module('extensions.sd-webui-controlnet.scripts.external_code', 'external_code')
+        self.assertEqual(self.ControlNetUnit, external_code_copy.ControlNetUnit)
+
+
+if __name__ == '__main__':
```
huchenlei, line 23: Tests looks good!

huchenlei (approved):
LGTM!

Thanks for cleaning it up.

scripts/external_code.py
ljleb: Yes, that is the idea. SD-CN-animation, deforum and other extensions import this module at this path so we cannot remove it completely. I guess a comment in the code would make this clear

PR #1778, opened 2023-07-08
compare with None because image may be array-like

aria1th:
fixes some external scripts such as https://github.com/ArtVentureX/sd-webui-agent-scheduler/


huchenlei:
So the current impl converts empty list `[]` to `None`, while the new impl will leave the value as empty list `[]`. Please also explain why this is important, i.e. what feature is depending on this behaviour.

aria1th:
@huchenlei Sorry for this, I found that it was problem in extension side, not your side. It was trying to directly set `unit['image']` as np.nadarray, whilst the required argument was `{'image': ndarray, 'mask' : ndarray | None}`.

PR #1798, opened 2023-07-18
API documentation improvements

langet0695:
I noticed that the old txt2img and img2img examples were out of date. 

To improve this removed the out of date documentation and replaced them with an updated example that illustrates to users how to send a request that uses the current syntax.

Hopefully this helps save others sometime as they ramp into the project.

Please let me know if you have any feed back.


huchenlei:
LGTM

Issue #1811, opened 2023-07-21
[Feature]: txt2img outpainting feathering/mask blur


catboxanon:
See https://github.com/Mikubill/sd-webui-controlnet/discussions/1597#discussioncomment-6401743 for background. 

[txt2img outpainting](https://github.com/Mikubill/sd-webui-controlnet/discussions/1464#discussioncomment-6054436) currently still produces a visible seam if you look closely enough. Depending on the image this can be more or less obvious in some cases. [The solution here](https://github.com/Mikubill/sd-webui-controlnet/discussions/1597#discussioncomment-6413756) is also not ideal because it's still a hard cut off, it still fails in some cases, and removes more of the original image that would be kept if feathering were used. The simplest method I see to work around this is employ a feathering technique, like the one already implemented in the webui's img2img inpaint functionality.

The idea would be to allow the inpainted area to be slightly larger by a given amount, as to replace part of the original image, but before the result is returned to the user, the original image would be composited (or "overlaid") on top and have the edges feathered by that same given amount. This is essentially the same thing the webui does for it's inpaint implementation from my understanding. 

https://github.com/AUTOMATIC1111/stable-diffusion-webui/blob/f865d3e11647dfd6c7b2cdf90dde24680e58acd8/modules/processing.py#L1196-L1206
https://github.com/AUTOMATIC1111/stable-diffusion-webui/blob/f865d3e11647dfd6c7b2cdf90dde24680e58acd8/modules/img2img.py#L117-L125

I plan to try implementing and opening a PR for this myself, but if this issue remains dormant for a few days you can safely assume I did not succeed in my attempt. 😅

PR #1813, opened 2023-07-21
Controllable inpainting Gaussian blur sigma value

catboxanon:
Closes https://github.com/Mikubill/sd-webui-controlnet/issues/1811

This exposes the Gaussian blur sigma value used for inpainting (currently hardcoded at an arbitrary value of `7`), which can allow for a more smoother transition in some cases, especially for txt2img outpainting. The default value remains at `7`.

You can see a before/after comparison here: https://imgsli.com/MTkzMzYw
(image from https://github.com/Mikubill/sd-webui-controlnet/discussions/1597#discussioncomment-6401743)

---

I believe this could still be improved but I'm not sure the best way to go about implementing a proper fix. An additional final pass in img2img + manual feathering may be the only way to 100% resolve it for the time being.

cc @wywywywy if you would like to test this out as well.

### DIFF
--- a/scripts/controlnet.py
+++ b/scripts/controlnet.py
@@ -901,7 +901,7 @@ class Script(scripts.Script, metaclass=(
                 final_inpaint_feed = np.ascontiguousarray(final_inpaint_feed).copy()
                 final_inpaint_mask = final_inpaint_feed[0, 3, :, :].astype(np.float32)
                 final_inpaint_raw = final_inpaint_feed[0, :3].astype(np.float32)
-                sigma = 7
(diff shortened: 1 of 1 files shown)


huchenlei:
LGTM on moving hard-coded number 7 to config. 👍

wywywywy:
Just tested it, and this works very well. Thank you so much.

Value of 30+ seems optimal for outpainting, while for actual inpainting 10 is pretty good.

So what would make it even better, is for this to become a generation parameter in my opinion.

image

aulerius:
Is it possible to experiment and overwrite this parameter through API calls?
I'm always having visible seams when outpointing via API manually using inpaint_only.

Issue #1802, opened 2023-07-19
I'm using adetailer which uses controlnet inpaint. Updated to latest version of controlnet and the inpaint is bugged


dilectiogames:
sorry. so the issue is described here with examples. I don't know which has the bug but same adetailer version with a previous controlnet version worked fine

https://github.com/Bing-su/adetailer/issues/216

huchenlei:
Not reproducable in my local setup. You can do a bisect between the version you upgraded from and the current version. If you can identify a culprit commit, we can see whether we can fix or revert it. Please re-open after you have identified culprit.

huchenlei:
Note: the core python code has not been updated for about a month. All recent changes are either front-end only, API change and documentation.

dilectiogames:
I've did some tests the version on controlnet v1.1.231 1770 dd766de8629ee6035a734217e08c26cd1b08b2ab) is the one that works perfectly and both v1.1.232 and v1.1.233 do not work. (I've tested the last commit of each version)

here is an image that shows the issue there are no artefacts in the hair, the eyelashes look more defined, the colors match the rest of the image, the body. this for the version v1.1.231

![example](https://github.com/Mikubill/sd-webui-controlnet/assets/126326386/c62be540-69c8-4df7-b63f-3442458d3e36)

if you want to replicate this you need

adetailer v23.6.3 and sd-webui-lora-block-weight 8ba434ef1e90217b8cd7c877e42980d3d1b4f9d6 all to last version as for today. I've only reverted controlnet to the previopus versions like in the image

you can get the prompts and the loras I've used form civitai from the the two links bellow. Save image as jpg and load in the png info to get the full prompt for adetailer. loras are on the right

https://civitai.com/images/1611646?period=AllTime&periodMode=published&sort=Newest&view=feed&username=dilectiogames&withTags=true

https://civitai.com/images/1601542?period=AllTime&periodMode=published&sort=Newest&view=feed&username=dilectiogames&withTags=true

thank you for your help and for the great work done on this extension hope this issue gets resolved. :D

dilectiogames:
I don't know how to reopen so hope you see the message above :)

huchenlei:
1.1.232 removes dynamic reloads of modules #1775, as in A1111 1.4.0 the extension reloading behaviour has been changed. Can you try to completely restart WebUI from commandline and see if the issue persists?

dilectiogames:
[...]

PR #1817, opened 2023-07-23
fix: do not assume images in postprocess_batch is a tensor

ljleb:
In sd-webui-comfyui we need `images` to be a list, so that each image in the list can have an arbitrary size, and so that the list size can change arbitrarily. It seems that `len(tensor)` is defined and equivalent to `tensor.shape[0]`.

See https://github.com/AUTOMATIC1111/stable-diffusion-webui/pull/11933 for details.


ljleb:
It seems that `len(tensor)` has been a thing for a long time:

https://github.com/pytorch/pytorch/blame/e3539a0e5497677a0f7ef29dc8278d249bbccb43/torch/_tensor.py#L907

This should then work no matter which version of python is installed.

huchenlei (approved):
LGTM

ljleb:
Thanks!

huchenlei:
Np. The PR passes web tests so I am confident that it does not break anything major. Also, can you take a look at #1802? I am not very familiar with ADetailer, but from the source code ADetailer is not trying to import anything directly from ControlNet. Why the behaviour changes after reload got removed?

ljleb:
Sure I'll look into it.

Issue #1801, opened 2023-07-19
[Bug]: Does not work with 16-bit depthmaps


semjon00:
### Is there an existing issue for this?

- [X] I have searched the existing issues and checked the recent builds/commits of both this extension and the webui

### What happened?
Depthmaps are routinely stored in 16-bit monochrome single-channel images. However, currently 16-bit depthmaps can not be used for the depth preprocessor/control. When you try to use it, Gradio will try to convert it to RGB. Sadly, this "conversion" simply truncates values into [0;255]. This is incorrect, the values should have been scaled down.

This issue belongs here since 16-bit depthmaps are reasonably popular, but WebUI probably won't ship updated Gradio for some time - even if this issue is fixed on the Gradio side.

Please see https://github.com/thygate/stable-diffusion-webui-depthmap-script/issues/291

### Steps to reproduce the problem

1. Supply 16-bit depthmap to "Single Image"
2. Click "Run preprocessor" (this is an easy way to reproduce the issue)
3. Preview gives away that the image was not correctly read.

### What should have happened?

16-bit depthmap works in exactly the same way as 8-bit depthmap. I think that this is ok if some precision is sacrificed in this specific usecase.

### Commit where the problem happens

webui: 1.4.0
controlnet: 1.1.233

### What browsers do you use to access the UI ?

Mozilla Firefox

### Command Line Arguments

```Shell
None
```

### List of enabled extensions

Only controlnet

### Console logs

```Shell
Nothing useful there
```

thomaseding:
You do realize the depth models are not trained on 16-bit depthmaps. It's not like the model can understand it in the first place without converting it to RGB beforehand. (And if I'm not mistaken, doesn't ControlNet put the control image into latent space?) Simply put, the model isn't nuanced enough to even understand the data.

thomaseding:
Please provide such a depth bitmap and I will take a look into it.

semjon00:
![mushroom-photography-brown-slimy-waxcap-hygrocybe-irrigata](https://github.com/Mikubill/sd-webui-controlnet/assets/54073010/2d4519eb-eb45-412d-94b3-2181e5579e0a)
![depthmap-16900988530003](https://github.com/Mikubill/sd-webui-controlnet/assets/54073010/9de422e3-9055-4de4-97e2-9557b44ea4db)

semjon00:
[...]

PR #1819, opened 2023-07-23
Fix 16-bit grayscale control image conversion.

thomaseding:
Resolves this issue https://github.com/Mikubill/sd-webui-controlnet/issues/1801

### DIFF
--- a/scripts/utils.py
+++ b/scripts/utils.py
@@ -3,12 +3,16 @@ import os
 import functools
 import time
 import base64
+import gradio as gr
 import numpy as np
 import safetensors.torch
 import logging
+import io
 
 from typing import Any, Callable, Dict
 from modules.safe import unsafe_torch_load
+
+from PIL import Image
 from scripts.logging import logger
 
 
@@ -120,7 +124,7 @@ except ImportError:
     pass
 
 
-def svg_preprocess(inputs: Dict, preprocess: Callable):
+def svg_preprocess(inputs: Dict):
     if not inputs:
         return None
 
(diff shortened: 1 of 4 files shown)


semjon00:
@thomaseding Thank you for this! Indeed, now this works with leres. Sadly, this commit does not completely solve the issue - the depthmap is still not read correctly if depth_midas is used as a preprocessor (maybe with others too, did not test them).

thomaseding:
Oh, I only tried it with the preprocessor as "None". I'll look more at it soon.

thomaseding:
@semjon00 I updated the code with a few bugfixes. Please try again.

semjon00:
@thomaseding Midas still does not work :(

thomaseding:
@semjon00 Can you provide an example and reproduction steps. Midas works on my end (I tried on two different computes). Here is a screenshot with Midas preprossor:
midas-example

semjon00:
Indeed, the mushroom image works flawlessly on my machine. Here is another image that I believed had exactly the same format, but now I see that it behaves differently from the mushroom image (this image creates problems).

depthmap-0424

Sorry, I really should have included it in my previous comment.

Exact repro steps <done 10 minutes ago>:
Installed extension from "Install from URL" (picked the correct branch), stopped WebUI, put control_11f1p_sd15_depth into the models folder, started WebUI, went to txt2img tab, clicked "Enabled" button in the ControlNet, then clicked the "Depth" control type, then drag-and-dropped the image above, then clicked the run preprocessor. Result: 

Untitled


Midas result if the same image is converted to jpeg:

Input:
depthmap-0424

Output:
image

These are visibly different, although I would expect them to be very similar - in every pixel, the jpg input is probably not more than 1/256 different from the png input (maybe they should actually look different?). Is it just midas being very bad for some reason, or is this another issue? I am quite puzzled now...

Commit 63423efb
WebUI v1.4.1

thomaseding:
@semjon00 The difference is due to JPEG compression. I took the 16-bit image and converted it to the following formats:
- 8 bit rgb (per channel)
- 8 bit gray
- 8 bit gray alpha
- 16 bit gray
- 16 bit gray alpha

And these all yielded the same output from Leres. 

Although we can't see the difference, it's enough for the Leres preprocessor (which is an AI model independent of SD) to pick up on.

semjon00:
Right, forgot for a second that JPEG is a lossy format. Silly me.
Indeed, changing the net size (Preprocessor Resolution) made it work perfectly! This fixes all the problems on my end.
Thank you again for the effort, now this extension would integrate well with the [Depth Maps script](https://github.com/thygate/stable-diffusion-webui-depthmap-script) and other places where one might get high-precision depthmaps. This makes me happy :)

About the net size - do you think the default size for midas could be changed to 384? I believe this may lead to better results. As far as I could understand, 384 is the suggested size for Midas v3.0 DTP-L (the model used). The UI could automatically set net size on model selection - 512 for zoedepth, 384 for midas, 448 for leres. Do you think it is a good idea; should I implement that?

thomaseding:
@semjon00 I have no idea about the defaults. All I know is that the code to change the defaults is trivial if you want to make a PR for it. All you need to do is change the corresponding entry in this table `preprocessor_sliders_config` in `scripts/processor.py`

scripts/utils.py
```diff
@@ -138,7 +140,82 @@ def svg_preprocess(inputs: Dict, preprocess: Callable):
...
+
+    def go_image(image, input):
+        if image is None:
+            return None
+        # gr.Image.preprocess() throws away depth information, so we need to decode again.
+        # Naively just checking if any elem is greater than 255 is not enough, because
+        # the image may be a (very dark) 16 bit grayscale image with all values below 256.
+        png_encoded = gr.processing_utils.extract_base64_data(input)
+        png = Image.open(io.BytesIO(base64.b64decode(png_encoded)))
+        assert len(png.tile) == 1
```
thomaseding, line 177: I should probably make this an `if` instead of an assert, in case the user sends an image stored with multiple tiles (probably very rare?). In that case, it's probably better that the image gets decoded without this scaling in case some of the other tiles don't correspond to the same depth as `tile[0]`

Either that or I learn how to handle tiles and scale the data appropriately. Might be as simple as sliding non-overlapping windows... which is quite doable.

scripts/utils.py
```diff
@@ -138,7 +140,82 @@ def svg_preprocess(inputs: Dict, preprocess: Callable):
         base64_str = str(encoded_string, "utf-8")
         base64_str = "data:image/png;base64," + base64_str
         inputs["image"] = base64_str
-    return preprocess(inputs)
```
huchenlei, line 141: So we are essentially overriding `gr.Image.preprocess`?

huchenlei (commented):
Thanks for you effort trying to fix this!

scripts/utils.py
thomaseding: Yeah. The new code intercepts the conversion to "RGB" and to "numpy" by changing the member variables before calling the "base" method. Then after my custom processing, I put it back to "RGB" and "numpy" afterward

scripts/utils.py
```diff
@@ -138,7 +140,82 @@ def svg_preprocess(inputs: Dict, preprocess: Callable):
...
+    return inputs
+
+
+# This is a patch for `gr.Image.preprocess()`
+def image_preprocess(self, x: Dict):
```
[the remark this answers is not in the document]
thomaseding: Sure, I'll look into it sometime this week. I haven't done GUI work in years, so I will have to play around a little.

scripts/utils.py
```diff
@@ -138,7 +140,82 @@ def svg_preprocess(inputs: Dict, preprocess: Callable):
...
+            "I;16N": 0x10000,
+        }
+        depth = mode_to_oversized_depth.get(data_mode, 256)
+        if depth > 256:
+            ratio = 256 / depth
+            data = np.array(image)
+            data = data * ratio
+            data = np.clip(data, 0, 255)
+            data = data.astype(np.uint8)
+            image = Image.fromarray(data, mode="L")
```
missionfloyd, line 195:
```suggestion
            image = image.point(lambda p: p * (256 / depth))
```
This can be done in one line.
https://pillow.readthedocs.io/en/stable/reference/Image.html#PIL.Image.Image.point

semjon00:
@thomaseding @missionfloyd This apparently has gone stale, bug the bugfix is still needed and quite important. @missionfloyd I would appreciate if this is merged as-is, as long as there are no critical blockers.

thomaseding:
@semjon00 Sorry, I haven't had much time to work on code outside of work. I probably won't have time to look at how to create tests for this until perhaps sometime in fall.

semjon00:
@huchenlei This is old, but still it would be nice to have it merged (as long as there are no critical issues).

PR #1828, opened 2023-07-25
update model_list to allow to set whether to update model

LeonZhao28:
add a parameter in the method `/controlnet/model_list` to allow user to set whether to update model

### DIFF
--- a/scripts/api.py
+++ b/scripts/api.py
@@ -33,7 +33,7 @@ def controlnet_api(_: gr.Blocks, app: FastAPI):
         return {"version": external_code.get_api_version()}
 
     @app.get("/controlnet/model_list")
-    async def model_list(update=True):
+    async def model_list(update: bool = True):
         up_to_date_model_list = external_code.get_models(update=update)
         logger.debug(up_to_date_model_list)
         return {"model_list": up_to_date_model_list}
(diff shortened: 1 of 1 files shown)


huchenlei:
Generally, a get request should not have side effect on the server. If you want an option to refresh model_list using web API, I would suggest adding a new POST endpoint.

LionelZhao28:
> Generally, a GET request should not have side effect on the server. If you want an option to refresh model_list using web API, I would suggest adding a new POST endpoint. 

Yes, I agree. But I actually hope to not reload the local model list every time I call this api. The change I pushed is actually to ensure consistency with existing Api usage. From an interface design perspective, it is not reasonable to force a refresh of the local model list. So, do you think it's better to keep the current habit of not changing the small-scale optimization API, or to make it more reasonable but lacking it will affect the usage of the existing API?

huchenlei:
Sorry I missed the context. Currently we are doing the refresh everytime the API is called. From that perspective, I think your change is reasonable. We should not break the API consistency without collecting metrics, as users might depend on existing behaviour.

Thanks for spotting the issue and fix it.

huchenlei (approved):
LGTM

PR #1832, opened 2023-07-26
Add setting to disable ControlNet in highres fix pass

### LINKED ISSUE #1826: [Feature Request] Bypass ControlNet on highres-fix second pass
ControlNet is a great tool for "initial shape" generation, but it has pretty big negative impact on visual quality of final image.

It would be great if I could disabled ControlNet on second pass of generation with highres-fix, so that StableDiffusion could fix all image issues in small details.

I saw previous issues on that feature-request but non of them worked for me in newer versions of ControlNet. Could you please add checkbox that disables controlnet on highres-fix second pass?

catboxanon:
Closes #1826

Adds an option in settings to disable ControlNet in highres fix. I'm not 100% sure if this breaks anything so I'm setting this as a draft for now. I also cannot produce anything that causes a "big negative impact" as mentioned in the issue, as in my tests images look nearly the same with or without this setting enabled, but that does mean it can cut down generation times in half with minimal change to the final output, so I figure it's a nice QoL feature to have.


huchenlei (approved):
Change LGTM. Let me know when it is ready for merge.

PR #1833, opened 2023-07-26
Repair ControlNet infotext parsing

### LINKED ISSUE #1793: [Bug]: PNG info, not sending controlnet params

### What happened?

Same issue as stated in:
#1203 

PNG info not sending ControlNet paramaters with  "Send to txt2img"

1. Produce image with using controlnet
2. Load produced image in PNG info
3. From PNG info, Send to txt2img/img2img 

-

webui: v1.4.1
controlnet: v1.1.232

Google Chrome

### Command Line Arguments

```Shell
--xformers --allow-code --api --api-log --autolaunch --enable-insecure-extension-access --administrator
```

### List of enabled extensions

-

### Console logs

```Shell
-
```

### Additional information

_No response_

catboxanon:
Closes #1793

This was broken (and has been for quite some time now) after ControlNet params started being stored altogether as part of a single key and not separate keys per param. The ideal solution would be rework how this is handled altogether so it accounts for more parameters in a cleaner way, but this is just patching the functionality that previously existed for now.

This also allows the `Parse ControlNet parameters in infotext` setting (previously titled `Passing ControlNet parameters with "Send to img2img"` which is unclear and technically incorrect since it also applies to txt2img) to be handled at runtime rather than requiring a restart of the webui, as the setting previously was only taken into account during the UI creation process.

~~Setting as a draft because [`active_units.js`](https://github.com/Mikubill/sd-webui-controlnet/blob/main/javascript/active_units.js) doesn't trigger UI updates when the infotext is parsed currently. Not sure what's going on as it does recognize the changes once the ControlNet tab is switched or the `Enabled` checkbox is unticked (easier to see this behavior if loading the infotext from an image using multiple ControlNet modules).~~

### DIFF
--- a/javascript/active_units.js
+++ b/javascript/active_units.js
@@ -6,7 +6,6 @@
  * Disable resize mode selection when A1111 img2img input is used.
  */
 (function () {
-    const cnetAllUnits = new Map/* <Element, ControlNetUnitTab> */();
     const cnetAllAccordions = new Set();
     onUiUpdate(() => {
         const ImgChangeType = {
@@ -64,8 +63,9 @@
         }
 
         class ControlNetUnitTab {
-            constructor(tab) {
+            constructor(tab, accordion) {
                 this.tab = tab;
+                this.accordion = accordion;
                 this.isImg2Img = tab.querySelector('.cnet-unit-enabled').id.includes('img2img');
 
                 this.enabledCheckbox = tab.querySelector('.cnet-unit-enabled input');
@@ -84,6 +84,7 @@
                 this.attachTabNavChangeObserver();
                 this.attachImageUploadListener();
                 this.attachImageStateChangeObserver();
+                this.attachA1111SendInfoObserver();
 
                 // Initial updates:
                 if (this.isImg2Img)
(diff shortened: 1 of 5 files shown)


huchenlei:
Active units is listening enable checkbox change and tab change events to update tab status. Setting checkbox value directly from scripts cannot be observed by js. A workaround would be add a listener to those send to buttons.

huchenlei:
Hey @catboxanon, I finally got some time to fix this issue. Can you confirm that the PR is good after my change?

huchenlei:
I think there is a big blocker for values to be pasted from infotext:

When module(preprocessor) value is updated, an update will be triggered to set the values of all sliders to the default value for the selected module. This behaviour will overwrite the pasted value from infotext for those sliders.

huchenlei:
The slider issue is very hard to fix. I think gradio has some hard time handling these concurrent updates.

I tried the following approach and it does not work:
- Make 2 invisible gradio buttons `build slider` and `build slider keep value`
- Use front-end `MutationObserver` to distinguish user operation on dropdown and Gradio update on the dropdown
- Dispatch user update to click `build slider`, and dispatch Gradio update (From A1111 paste) to `build slider keep value`

However, when A1111 paste the infotext to ControlNetUnit, when `build slider keep value` call is reached in backend python code, the state of module and pp are not necessarily up to date.

PR #1875, opened 2023-08-05
reset repo to 5ae9b4a

lllyasviel:
This reset repo to 5ae9b4a1a0c7d9a2938e75aaf052ab078623066f.

Someone informed me that the installation on windows is broken.

I tested with 4 windows machines and 1 ubuntu machine. All windows failed. The ubuntu succeeded but spent one minute to install everything.

Note that mmcv and detectron are on the dangerous list that we should never add to the dependency. If we must, we need to rewrite it in the same way with https://github.com/Mikubill/sd-webui-controlnet/tree/main/annotator/mmpkg/mmseg to avoid any installation that requires compile.

And we should never include anything that requires compile.


petercham:
自动更新后现在版本还是1.1.234，是回退版本了么？

huchenlei:
> 自动更新后现在版本还是1.1.234，是回退版本了么？

我把reset之前的版本发到了`dw_pose_reland`分支，你仍然可以通过switch到这个分支来使用dw pose的功能。我们正在解决dependency需要compile cpp的问题。 @petercham

huchenlei:
@lllyasviel So what is your opinion on annotating the training data with this new annotator and re-train the ControlNet openpose model with feet support and better hand control?

lllyasviel:
Well. I am aware of these new models since several (>4) months ago. Like this
https://github.com/IDEA-Research/HumanArt
image

And the DWPose is another option.

The main workload is to rewrite mmcv so that it can work outside the terrible openmim compile and dependency. 

Note that, many months ago, we were in the same situation to port Uniformer segmentation model from mmcv’s mmseg. We succeeded in extracting the complie-free codes - that is just the current seg_ufade20k. This took us efforts of about two weeks: https://github.com/Mikubill/sd-webui-controlnet/tree/main/annotator/mmpkg 

I think is it possible to make it works if we spend another two weeks on mmpose to extract it, but I do not think I have enough free time to handle that.

huchenlei:
Do you mean that the ControlNet v1.1. openpose model is already using very accuratly annotaed openpose skeletons in the training dataset?

I will take a look at how to port mmpose without build dependencies on cpp.

Dasor92:
i was on windows 11 and everything worked fine on 397f54ba3148dfcdf9f3cd455b705f7595487485 what was the problem? i even used DWPose several times to test it out

yzd-v:
We will release a new DWPose version with onnx. It will not rely on any mmxx. This avoids rewriting mmcv.

huchenlei:
> We will release a new DWPose version with onnx. It will not rely on any mmxx. This avoids rewriting mmcv.

Well, Onnx still need MS Visual C++ installation on windows. See https://onnxruntime.ai/docs/install/#requirements 

I am assuming that it also needs to compile something.

WDYT? @yzd-v @lllyasviel

huchenlei:
I have made `onnxruntime` an optional dependency in #1892.

PR #1882, opened 2023-08-06
opencv-python-headless

tcmaps:
This replaces the opencv package with a headless variant, to avoid dependency on libglib2, libgl1 etc. Ref #1881


tcmaps:
See Dockerfile: https://github.com/Mikubill/sd-webui-controlnet/blob/af34f514499933d5e7e1641a6b13d56411e45e76/annotator/zoe/zoedepth/models/base_models/midas_repo/Dockerfile#L4

Issue #1867, opened 2023-08-04
Preview not showing if img2img image is being used


NoteToSelfFindGoodNickname:
If you don't load an image into ControlNet, it will show a small green button showing "AI1111", and when you hover over it, it says "If you don't load an image here, the img2img will be used".

That is great. It does show the img2img2 image, I don't have it load it.

However, when I click the Explode button, no preview is shown.
Only when I load the image manually, the Explode button shows a preview.

huchenlei:
See note of https://github.com/Mikubill/sd-webui-controlnet/pull/1747. 

I was trying to display an alert when there you click the run processor button without an input image, but obviously Gradio does not like vanilla JS `alert` which is blocking.

I will try to implement it another way. Thanks for reminding!

Trigger the preprocessor with the A1111 image directly is not quite feasible as the A1111 image has many sources (inpaint, inpaint upload, img2img, etc)

PR #1887, opened 2023-08-07
disable run preprocessor if no cn input image

Guillaume-Fgt:
In response to issue https://github.com/Mikubill/sd-webui-controlnet/issues/1867 and related to PR [#1747](https://github.com/Mikubill/sd-webui-controlnet/pull/1747), I suggest those modifications to the Javascript code.

The run preprocessor button is enabled only if an image input exists in the ControlNet input tab. The hint of the button is changed accordingly.

One problem I have:
at the first start of webui, the button is enabled without any input image. I don't know where in the code I can change this initial state. But I don't think it is a big issue as the user will not expect any result with a totally blank canvas I believe.


huchenlei:
I moved the detection logic to `active_units.js` as this logic should work for both txt2img and img2img tabs.

PR #1892, opened 2023-08-07
Reland Reland "Port DW Pose preprocessor"

huchenlei:
Relands https://github.com/Mikubill/sd-webui-controlnet/pull/1890.

This PR now makes onnx runtime optional. (Not forcing user to install it on startup)

PR #1897, opened 2023-08-09
:bug: Add dependency on opencv 4.8.0

huchenlei:
User has reported that opencv 4.5.5 cannot run onnx model. https://github.com/Mikubill/sd-webui-controlnet/pull/1892#issuecomment-1671727728

This PR adds the opencv dependency with a known good version.

### DIFF
--- a/install.py
+++ b/install.py
@@ -1,8 +1,21 @@
 import launch
 import os
 import pkg_resources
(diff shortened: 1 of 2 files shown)


Chinczyk666:
H

PR #1898, opened 2023-08-09
Revert ":bug: Add dependency on opencv 4.8.0"

huchenlei:
Reverts Mikubill/sd-webui-controlnet#1897

pip install opencv-python>=4.8.0 does not perform the package upgrade correctly in local testing.

### DIFF
--- a/install.py
+++ b/install.py
@@ -1,21 +1,8 @@
 import launch
 import os
 import pkg_resources
-from typing import Tuple, Optional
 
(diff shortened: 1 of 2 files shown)


SLAPaper:
maybe should use `pip install -U` instead ?

huchenlei:
@SLAPaper Thanks! Adding `-U` helps.

PR #1907, opened 2023-08-11
Fix mask blur

catboxanon:
Resolves https://github.com/Mikubill/sd-webui-controlnet/discussions/1888, https://github.com/AUTOMATIC1111/stable-diffusion-webui/issues/12428

This has been broken for about 3 months now and was never accounted for in this extension. The webui switched from using `mask_blur` to separate attributes for `mask_blur_x` and `mask_blur_y` in 1.4.0, so the `mask_blur` was read as `None` by ControlNet. https://github.com/AUTOMATIC1111/stable-diffusion-webui/pull/10295

Uses the same code from here to modify the mask: https://github.com/AUTOMATIC1111/stable-diffusion-webui/blob/26c92f056acc795af5066779f1b8aedb8dfa983d/modules/processing.py#L1287-L1297

### DIFF
--- a/scripts/controlnet.py
+++ b/scripts/controlnet.py
@@ -163,8 +163,22 @@ def prepare_mask(
     mask = mask.convert("L")
     if getattr(p, "inpainting_mask_invert", False):
         mask = ImageOps.invert(mask)
-    if getattr(p, "mask_blur", 0) > 0:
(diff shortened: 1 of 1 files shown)


huchenlei (approved):
LGTM

Issue #1933, opened 2023-08-19
[DevTask] Support Stability AI's Control-LoRA for SDXL


huchenlei:
Stability released the official ControlNet support for SDXL here:
https://huggingface.co/stabilityai/control-lora

Instead of 5GB original size model, they now offer option of ~800MB M size model and ~400MB S size model.

Stability's implementation in SwarmUI relies on Comfy's implementation here:
https://github.com/comfyanonymous/ComfyUI/commit/d6e4b342e6171e0c89fed1ec9998b608888af401

We should try to port the change here.

huchenlei:
I took some time studying Comfy's impl. It seems like the main change to support ControlNet LoRA is replacing all calls of `linear`, `conv_nd` with following custom impl that accounts `up` and `down` tensors.

```python
# linear forward
def forward(self, input):
            if self.up is not None:
                return torch.nn.functional.linear(input, self.weight + (torch.mm(self.up.flatten(start_dim=1), self.down.flatten(start_dim=1))).reshape(self.weight.shape).type(self.weight.dtype), self.bias)
            else:
                return torch.nn.functional.linear(input, self.weight, self.bias)
```

```python
# conv_2d forward
def forward(self, input):
            if self.up is not None:
                return torch.nn.functional.conv2d(input, self.weight + (torch.mm(self.up.flatten(start_dim=1), self.down.flatten(start_dim=1))).reshape(self.weight.shape).type(self.weight.dtype), self.bias, self.stride, self.padding, self.dilation, self.groups)
            else:
                return torch.nn.functional.conv2d(input, self.weight, self.bias, self.stride, self.padding, self.dilation, self.groups)
```

Those callsites are not only in Comfy's ControlNet specific code, but also `ldm` code. I am not sure how should we handle that here, maybe we should also have a modified copy of ldm code? @lllyasviel

lllyasviel:
> I took some time studying Comfy's impl. It seems like the main change to support ControlNet LoRA is replacing all calls of `linear`, `conv_nd` with following custom impl that accounts `up` and `down` tensors.
> 
> ```python
> # linear forward
> def forward(self, input):
>             if self.up is not None:
>                 return torch.nn.functional.linear(input, self.weight + (torch.mm(self.up.flatten(start_dim=1), self.down.flatten(start_dim=1))).reshape(self.weight.shape).type(self.weight.dtype), self.bias)
>             else:
[...]

PR #1936, opened 2023-08-21
Support ControlNet LoRA

huchenlei:
This PR ports following change in ComfyUI(https://github.com/comfyanonymous/ComfyUI/commit/d6e4b342e6171e0c89fed1ec9998b608888af401).

There are seveal things this PR is going to achieve:

1. Correctly recognize ControlNet LoRA and convert its `state_dict` to acceptable format
2. Use custom `linear`, `conv_nd` operations when a ControlNet LoRA is loaded
3. Create yaml configs for ControlNet LoRA models

Questions to be answered:
1. Will this approach support composing multiple ControlNet LoRAs?
2. Some code is ported from an GPL license repo (Comfy). Probably need to either rewrite / contact Comfy for license conflict solution.

### DIFF
--- a/scripts/hook.py
+++ b/scripts/hook.py
@@ -46,11 +46,17 @@ def mark_prompt_context(x, positive):
         return x
     if isinstance(x, ScheduledPromptConditioning):
         cond = x.cond
-        if prompt_context_is_marked(cond):
+        tensor_cond = cond if isinstance(cond, torch.Tensor) else cond['crossattn']
+        if prompt_context_is_marked(tensor_cond):
             return x
         mark = POSITIVE_MARK_TOKEN if positive else NEGATIVE_MARK_TOKEN
-        cond = torch.cat([torch.zeros_like(cond)[:1] + mark, cond], dim=0)
-        return ScheduledPromptConditioning(end_at_step=x.end_at_step, cond=cond)
+        tensor_cond = torch.cat([torch.zeros_like(tensor_cond)[:1] + mark, tensor_cond], dim=0)
+
+        return ScheduledPromptConditioning(end_at_step=x.end_at_step, 
+                                           cond=(tensor_cond if isinstance(cond, torch.Tensor) else {
+                                               **cond,
+                                               'crossattn': tensor_cond,
+                                           }))
     return x
 
 
@@ -334,6 +340,8 @@ class UnetHook(nn.Module):
             param.guidance_stopped = current_sampling_percent < param.start_guidance_percent or current_sampling_percent > param.stop_guidance_percent
 
     def hook(self, model, sd_ldm, control_params, process):
+        is_sdxl = getattr(process.sd_model, 'is_sdxl', False)
+
         self.model = model
         self.sd_ldm = sd_ldm
         self.control_params = control_params
@@ -355,8 +363,6 @@ class UnetHook(nn.Module):
             return process.sample_before_CN_hack(*args, **kwargs)
 
         def forward(self, x, timesteps=None, context=None, **kwargs):
-            total_controlnet_embedding = [0.0] * 13
-            total_t2i_adapter_embedding = [0.0] * 4
             require_inpaint_hijack = False
             is_in_high_res_fix = False
             batch_size = int(x.shape[0])
@@ -423,6 +429,11 @@ class UnetHook(nn.Module):
                 context = torch.cat([context, control.clone()], dim=1)
 
             # handle ControlNet / T2I_Adapter
+            input_block_num = 9 if is_sdxl else 12
+            assert input_block_num % 3 == 0
+            total_controlnet_embedding = [0.0] * (input_block_num + 1)
+            total_t2i_adapter_embedding = [0.0] * (input_block_num // 3 + 1)
+
             for param in outer.control_params:
                 if no_high_res_control:
                     continue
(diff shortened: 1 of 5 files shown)


scripts/hook.py
```diff
@@ -763,7 +769,7 @@ def hacked_group_norm_forward(self, *args, **kwargs):
         gn_modules = [model.middle_block]
         model.middle_block.gn_weight = 0
 
-        input_block_indices = [4, 5, 7, 8, 10, 11]
+        input_block_indices = [4, 5, 7, 8, 10, 11] if not getattr(process.sd_model, 'is_sdxl', False) else [4, 5, 7, 8]
```
huchenlei, line 772: SDXL only has 9 input blocks.

scripts/hook.py
```diff
@@ -600,8 +606,8 @@ def forward(self, x, timesteps=None, context=None, **kwargs):
             h = aligned_adding(h, total_controlnet_embedding.pop(), require_inpaint_hijack)
 
             # U-Net Decoder
-            for i, module in enumerate(self.output_blocks):
-                h = th.cat([h, aligned_adding(hs.pop(), total_controlnet_embedding.pop(), require_inpaint_hijack)], dim=1)
+            for module, (hs_item, controlnet_embedding_item) in zip(self.output_blocks, reversed(list(zip(hs, total_controlnet_embedding)))):
```
huchenlei, line 609: `hs` has 9 tensor elements, while `total_controlnet_embedding` has 10 tensor elements and 2 `0` padding at the end.

This change aligns the 2 arrays, but I am not sure it is the correct fix here.

huchenlei:
Currently got blocked on transformer arch issue. See details in https://github.com/Mikubill/sd-webui-controlnet/issues/1933#issuecomment-1685446217

In comfy's implementation, transformer depth is different at each level.

`label_emb` is another issue as there are weights for that field in the LoRA state dict.

huchenlei:
Ok, I have proved that the LoRA can run on A1111. Just there is a change necessary in ldm's `SpatialTransformer`. I will look to find a workaround that does not need to modify ldm source later.

Screen Capture 022 - Stable Diffusion - localhost

huchenlei:
Comfy UI has following line doing some extra adjustment on input `emb`
```python
        if self.num_classes is not None:
            assert y.shape[0] == x.shape[0]
            emb = emb + self.label_emb(y)
```

Where label_emb is a network that converts `adm` to `emb`'s shape.
```python
self.label_emb = nn.Sequential(
                    nn.Sequential(
                        linear(adm_in_channels, time_embed_dim, dtype=self.dtype),
                        nn.SiLU(),
                        linear(time_embed_dim, time_embed_dim, dtype=self.dtype),
                    )
                )
```

And `adm` in comfy is calculated differently for refiner and base SDXL models:
https://github.com/comfyanonymous/ComfyUI/blob/763b0cf024c8fd462343ab0a8cfdab099714168b/comfy/model_base.py#L157C1-L207

I am pretty sure something similar must also exists in A1111 and we just need to perform some mapping to get it to a format that `label_emb` network accepts.

From my testing, without the label_emb adjustment on emb, the result does not follow the ControlNet guide that well. I have to increase ControlNet weight to 1.5 to show that ControlNet is indeed guiding the generation.

scripts/hook.py
huchenlei: That impl does not handle a tailing 0 for middle block weights. Revised the impl to calculate array length based on model type.

huchenlei:
Some testings on rank128 size S models:

Canny:
tmprsib75w5


Depth:
tmp9zngl_yq


Re-color:
tmpw4aoqgxs


Using multiple units seems to be fine, but produce result with generally worse qualify. (Kinda expected?)

@lllyasviel PTAL

FurkanGozukara:
thank you so much amazing work

George0726:
thank you so much amazing and fast work

lllyasviel:
I need to take a look if there is something wrong in the impl before it is going to merge, the saturation seems a bit wired and I am not sure that some of those artifacts are limitations of the released models. Also, we should begin to merge those after webui 1.6.x since there are much differences between 1.5.x and 1.6.x

huchenlei:
> I need to take a look if there is something wrong in the impl before it is going to merge, the saturation seems a bit weird and I am not sure that some of those artifacts are limitations of the released models. Also, we should begin to merge those after webui 1.6.x since there are much differences between 1.5.x and 1.6.x

I tested the generations under 1.5.1 release tag and current HEAD of dev branch, which produces similar results.

I am not sure about the saturation issue. You can checkout the branch and do some testing yourself. My prompt is very simple here
`1man holding microphone`
`neg: abstract, black and white`

The comfy UI results I attached eariler is probably not comparable as the workflow they offer is img2img. Here are some canny img2img generation results with A1111:
grid-0000

lllyasviel:
where you find control-lora-sdxl.yaml?

lllyasviel:
wait. it seems that they are using a lora to construct a controlnet to control xl ...
why dont they directly use a lora to control xl?

FurkanGozukara:
> wait. it seems that they are using a lora to construct a controlnet to control xl ... why dont they directly use a lora to control xl?

i don't think so they are pro as you

so many people right now expecting controlnet for auto1111

lllyasviel:
to be continued in https://github.com/Mikubill/sd-webui-controlnet/pull/1952

huchenlei:
> where you find control-lora-sdxl.yaml?

I dumped the config out of Comfy and made that yaml.

PR #1941, opened 2023-08-22
remove annoying i2i preview

lllyasviel:
show image of a1111 in i2i preview is a good idea in theory, but in reality it is just annoying and misleading. take a look at issue and we are going to remove it

### DIFF
--- a/javascript/img2img_input_display.js
+++ b/javascript/img2img_input_display.js
@@ -4,6 +4,8 @@
  * that if no ControlNet input image is uploaded, by default ControlNet will
  * fallback onto img2img input image.
  */
+
+/*
 (function () {
     function getActiveImg2ImgTabImgSrc(img2imgTabs) {
         const tabs = img2imgTabs.querySelectorAll('.tabitem');
@@ -155,3 +157,4 @@
         callback_registered = true;
     });
 })();
+*/
--- a/scripts/controlnet_version.py
+++ b/scripts/controlnet_version.py
@@ -1,4 +1,4 @@
-version_flag = 'v1.1.301'
+version_flag = 'v1.1.302'
 
 from scripts.logging import logger
 


huchenlei:
Is there any replacement mechanism to let user know that A1111 input image will be used as fallback if no image is uploaded to ControlNet?

I implemented the preview because there are aboundant tutorial videos online re-upload the same image to ControlNet before, and some vidoes do suggest this fallback behaviour as "hidden trick".

lllyasviel:
> > Is there any replacement mechanism to let user know that A1111 input image will be used as fallback if no image is uploaded to ControlNet?
> 
> I'm seconding this. I don't really understand why this was removed. How is this "annoying and misleading"? It seems quite the opposite without this.

because it make users think that they have uploaded images but they are disappointed that they cannot click preprocessor preview. some even think the functionality of cn preview is broken, and sometimes the preview button no longer works.

the fallback to a1111 images should be a feature only for people who know it. We should not expect new users to know it. Just let new people to upload same images, and sometimes they will find the feature.

The trouble caused by this fake upload is way more significant than the benefits

lllyasviel:
it is also quite annoying when you can upload images without click gradio X button to remove existing images. People who know it will know this is becsuse that image is only a hint, not a real image, and this is ok. but for people who does not know it, this directly make cn image upload having two different behaviors (need or need not clicking x to remove existing images). This is just against every ui design principles and make things confusing and misleading

huchenlei:
Ok, so if the fallback placeholder image from A1111 img2img acts like an real ControlNet input, most these concerns should be solved right?

Previously that was not implemented as having a "real" ControlNet input image will make some masking logic break.

lllyasviel:
in theory it should be ok if it becomes a real uplpaded images, but if we do that we will probably trigger many gradio calls in many controlnet units.

my point is that the fallback to a1111 image should just be a feature for advanced users who know it. If users upload same images in i2i, it is just fine. we do not need to use UI to teach features.

PR #1970, opened 2023-08-24
Fix zoom and mask brush color customizable

w-e-w:
### make zoom work with single CN module 
currently it only works with when `Multi ControlNet: Max models amount` > 1

I took a shortcut in not checking if `window.applyZoomAndPan` is a function because 
 `window.applyZoomAndPanIntegration` is a new addition, and if it is a function then `window.applyZoomAndPan` also is function 

### make the mask brush color customizable
this is a new feature of gradio
and we have already applied it into 1.6.0 RC
which can be seen in the webui 1.6.0 RC or dev, but cannot be seen in master
I did a quick test and it seems that this doesn't break the current master 1.5.2, so hopefully it is backwards compatible with previous versions as well

I'm using the same settings key with img2img inpaint canvus in 1.6.0RC `shared.opts.img2img_inpaint_mask_brush_color`
but you may want to separate it out into a different key

screenshot of canvus zoom working with singel CN module and the brush color set to red
image

### DIFF
--- a/javascript/canvas.js
+++ b/javascript/canvas.js
@@ -6,6 +6,8 @@
                 hasApplied = true;
                 window.applyZoomAndPanIntegration("#txt2img_controlnet",Array.from({ length: 20 }, (_, i) => `#txt2img_controlnet_ControlNet-${i}_input_image`));
                 window.applyZoomAndPanIntegration("#img2img_controlnet",Array.from({ length: 20 }, (_, i) => `#img2img_controlnet_ControlNet-${i}_input_image`));
+                window.applyZoomAndPan("#txt2img_controlnet_ControlNet_input_image");
+                window.applyZoomAndPan("#img2img_controlnet_ControlNet_input_image");
                 //console.log("window.applyZoomAndPanIntegration applied.");
             } else {
                 //console.log("window.applyZoomAndPanIntegration is not available.");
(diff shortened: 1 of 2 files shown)


w-e-w:
odd, test passed in my repo test https://github.com/w-e-w/sd-webui-controlnet/actions/runs/5965562694 but failed here

lllyasviel:
probably do we add both functions to function test in that if.

w-e-w:
I found the cause

lllyasviel:
perhaps ci failed because cn is using old version of webui to do tests. old gradio does not support brush color. 
this is actually expected since a number of users do not update webui and cn is considered to be downward compatible

w-e-w:
> old gradio does not support brush color

since `gr.Image()` has `**kwargs` the worst that can happen is presented with an unknown parameter is that it doesn't know what to do but it doesn't throw an error
the error was caused by me calling `shared.opts.img2img_inpaint_mask_brush_color` which did not exist in previous versions so throwing an attribute error

> but should add window.applyZoomAndPan to that if line to avoid future changes in canvas zoom codebase

sure give me a moment

w-e-w:
I was trying to get the canvas zoom to also work on previous versions of webUI using `applyZoomAndPan` 1.4.0
before the introduction of `applyZoomAndPanIntegration` 1.6.0
but I couldn't, I even somehow managed to hang my browser

so this basically checks if the version is greater than 1.6.0 and have both functions before trying to apply

lllyasviel:
because this flow is a bit wired, i need to take a look at this again later

w-e-w:
feel free to make any changes you feel necessary

Issue #1854, opened 2023-08-02
[Bug]: Enable controlNet Units by default!


p0mad:
### Is there an existing issue for this?

- [X] I have searched the existing issues and checked the recent builds/commits of both this extension and the webui

### What happened?

How can i enable Controlnet Units by default without need to enable each one manually with their own configs!

Thanks
Regards

### Steps to reproduce the problem

N

### What should have happened?

N

### Commit where the problem happens

webui: version: [v1.5.1](https://github.com/AUTOMATIC1111/stable-diffusion-webui/commit/68f336bd994bed5442ad95bad6b6ad5564a5409a)  •  python: 3.10.6  •  torch: 2.0.1+cu118  •  xformers: 0.0.20  •  gradio: 3.32.0  •  checkpoint: [ef49fbb25f](https://google.com/search?q=ef49fbb25fa908bb54ded95abf81ef4b0e21fa0a8de56c40c3d62e768ef7e49a)
controlnet: Latest

### Command Line Arguments

```Shell
n
```

### List of enabled extensions

n

### Console logs

```Shell
n
```

huchenlei:
See https://github.com/Mikubill/sd-webui-controlnet/discussions/1703

p0mad:
@huchenlei 
Thanks for the reply, but i meant something else!
in AUTO1111 i can put my default configs inside ui-config.json to follow them as a default option!
but there is no such a thing for ControlNET!

I mean the default to be enabled, or some default setting for each unit of CN
that would be fascinating in case of using other script or testing same CN units multiple times ...
Thanks
Regards

huchenlei:
What do you mean `in case of using other scripts`? Are you expecting other scripts to manipulate ControlNet's UI? 

If you just want to do automation with some ControlNet presets, you should use the [Web API](https://github.com/Mikubill/sd-webui-controlnet/wiki/API) to achieve it.

p0mad:
@huchenlei 
by other Scripts, i mean using m2m, ...
I mean if such a thing is provided, that would kill lots of manual processing of doing each steps by hand!

Something like this but for CN:
"txt2img/Prompt/visible": true,
    "txt2img/Prompt/value": "",
    "txt2img/Negative prompt/visible": true,
    "txt2img/Negative prompt/value": "",
    "txt2img/Interrupt/visible": true,
    "txt2img/Skip/visible": true,
    ...

huchenlei:
[...]

PR #1974, opened 2023-08-24
[New Feature] ControlNet unit preset management

huchenlei:
Users have requested to be able to save easily save and load controlnet unit presets in https://github.com/Mikubill/sd-webui-controlnet/issues/1854.

This PR implements a preset manager UI that allows users to save/load ControlNet unit presets. Presets are persisted locally as text file under `presets/`.

 
Screen Capture 025 - Stable Diffusion - 127 0 0 1

Known issues:
- Preprocessor sliders values are not populated due to issue mentioned in #1833 .
- Toolbuttons are having wrong hover hint.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -174,3 +174,6 @@ annotator/downloads/
 web_tests/results/
 web_tests/expectations/
 *_diff.png
+
+# Presets
+presets/
--- a/javascript/active_units.js
+++ b/javascript/active_units.js
@@ -85,6 +85,7 @@
                 this.attachImageUploadListener();
                 this.attachImageStateChangeObserver();
                 this.attachA1111SendInfoObserver();
+                this.attachPresetDropdownObserver();
 
                 // Initial updates:
                 if (this.isImg2Img)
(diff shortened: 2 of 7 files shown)


lllyasviel:
will you maintain this?

huchenlei:
Hint fixed.

Sure I will be maintaining this part. It is not complex.

lllyasviel:
control type is not stored, make it feel a a bit unpleasant.
Also it should restore blank when controlnet_ui_group.py detects changes. if you get time you may take a look at line 843-872 in controlnet_ui_group.py to see how to implement this

huchenlei:
The design here is that if user make changes with a preset selected, pressing save button will overwrite that preset. To save as a new preset, user need to select new preset and click save.

Going to implement storage of control type later.

lllyasviel:
but if I selected present A and made some changes, how can I go back to A using the present A?
This should be the major use case of preset

huchenlei:
Yea, it can be awkward to select `new preset` and then select `preset A` to clear modifications made.

However, I believe there need to be a way to to edit the preset. Do you think it would be better if we make an edit button and let user edit the raw infostring in a popup box?

lllyasviel:
you may want to design it a bit - and if you think the current form is not mature enough this can be temporarily reverted.
but it is ok if just put here and update it if you move fast

PR #1993, opened 2023-08-26
Migrate tooltips to Gradio

catboxanon:
Title. This is actually already being done as well in https://github.com/Mikubill/sd-webui-controlnet/blob/main/scripts/controlnet_ui/preset.py, so it should be kept consistent. Also makes the frontend faster as a bonus since `querySelectorAll` is no longer being called on every UI update (these should really be optimized for [`modal.js`](https://github.com/Mikubill/sd-webui-controlnet/blob/eacfe995cf7ea7991e88e4bb4700a57adf081cbb/javascript/modal.js#L4-L8) and [`openpose_editor.js`](https://github.com/Mikubill/sd-webui-controlnet/blob/eacfe995cf7ea7991e88e4bb4700a57adf081cbb/javascript/openpose_editor.js#L103) as well because it makes the UI very sluggish, to the point I've blocked my browser from accessing these resources).


huchenlei:
LGTM.

huchenlei:
Part of warning reduction effort: https://github.com/Mikubill/sd-webui-controlnet/issues/2229

PR #2068, opened 2023-09-06
Fix for IP-Adapter preprocessor download

FNSpd:
Previously failed with this error:
ImportError: cannot import name 'load_file_from_url' from 'modules.modelloader'


lllyasviel:
cn 1.1.4 only support webui beyond 1.6.0

PR #2123, opened 2023-09-17
Update movie2movie.py

dsalvat1:
Bruh, remove the coma after duration


dsalvat1:
I realized it is an unpack, still it is having problems with duration 

/stable-diffusion-webui/extensions/sd-webui-controlnet/scripts/movie2movie.py", line 126, in run

duration, = args[contents_num * arg_num:]

ValueError: not enough values to unpack (expected 1, got 0)

PR #2139, opened 2023-09-22
Fix debug loglevel

huchenlei:
`process_has_sdxl_refiner` is a static method on `Script`. Calling it with `self.` prefix will cause issue when the method is decorated by

```python
def timer_decorator(func):
    """Time the decorated function and output the result to debug logger."""
    if logger.level != logging.DEBUG:
        return func

    @functools.wraps(func)
    def wrapper(*args, **kwargs):
        start_time = time.time()
        result = func(*args, **kwargs)
        end_time = time.time()
        duration = end_time - start_time
        # Only report function that are significant enough.
        if duration > 1e-3:
            logger.debug(f"{func.__name__} ran in: {duration} sec")
        return result

    return wrapper
```

in debug loglevel, as `self.process_has_sdxl_refiner(p)` will be interpreted as `process_has_sdxl_refiner(self, p)`, but the function only accepts single param.

This PR fixes the issue by replacing `self.` with `Script.` in all callsites.

### DIFF
--- a/scripts/controlnet.py
+++ b/scripts/controlnet.py
@@ -973,7 +973,7 @@ class Script(scripts.Script, metaclass=(
         return getattr(p, 'refiner_checkpoint', None) is not None
 
     def process(self, p, *args, **kwargs):
-        if not self.process_has_sdxl_refiner(p):
+        if not Script.process_has_sdxl_refiner(p):
             self.controlnet_hack(p)
         return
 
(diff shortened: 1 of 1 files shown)


huchenlei:
Sample stack trace:
```
*** Error running before_process_batch: D:\stable-diffusion-webui\extensions\sd-webui-controlnet\scripts\controlnet.py
      File "D:\stable-diffusion-webui\modules\scripts.py", line 631, in before_process_batch
        script.before_process_batch(p, *script_args, **kwargs)
      File "D:\stable-diffusion-webui\extensions\sd-webui-controlnet\scripts\utils.py", line 90, in wrapper
        result = func(*args, **kwargs)
      File "D:\stable-diffusion-webui\extensions\sd-webui-controlnet\scripts\controlnet.py", line 986, in before_process_batch
        if self.process_has_sdxl_refiner(p):
      File "D:\stable-diffusion-webui\extensions\sd-webui-controlnet\scripts\utils.py", line 90, in wrapper
        result = func(*args, **kwargs)
    TypeError: Script.process_has_sdxl_refiner() takes 1 positional argument but 2 were given
```

PR #2186, opened 2023-10-20
increase possible model cache size

MitPitt:
Good to have for larger pipelines


huchenlei:
LGTM. I am fine with the config change if that is necessary to fit users need.

PR #2248, opened 2023-11-10
Fix the issue of not recognizing the use CPU mode

kingljl:
Fix Controlnet cannot support the -- use CPU parameter, built-in will still load cuda

### DIFF
--- a/annotator/clipvision/__init__.py
+++ b/annotator/clipvision/__init__.py
@@ -79,10 +79,10 @@ downloads = {
 
 
 clip_vision_h_uc = os.path.join(os.path.dirname(os.path.abspath(__file__)), 'clip_vision_h_uc.data')
-clip_vision_h_uc = torch.load(clip_vision_h_uc)['uc']
+clip_vision_h_uc = torch.load(clip_vision_h_uc,  map_location=torch.device('cuda' if torch.cuda.is_available() else 'cpu'))['uc']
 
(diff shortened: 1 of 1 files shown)


huchenlei:
Can you add a unittest for this? All unittests run on CPU.

You can add a new test case in https://github.com/Mikubill/sd-webui-controlnet/blob/main/tests/web_api/txt2img_test.py, and add a new download line here:
https://github.com/Mikubill/sd-webui-controlnet/blob/05ef0b1cd1374cf285dd8d5ccd7db9997549893c/.github/workflows/tests.yml#L59-L61

kingljl:
> Can you add a unittest for this? All unittests run on CPU.
> 
> You can add a new test case in https://github.com/Mikubill/sd-webui-controlnet/blob/main/tests/web_api/txt2img_test.py, and add a new download line here:
> 
> https://github.com/Mikubill/sd-webui-controlnet/blob/05ef0b1cd1374cf285dd8d5ccd7db9997549893c/.github/workflows/tests.yml#L59-L61

It's like this, when Controlnet is running, if -- use CPU is specified, the program will still execute in cuda mode. Will prompt cuda.is_ Available error, unable to flexibly select. Here, I just added cuda and CPU judgment in torch. load. The original one side can run directly
As shown in the figure:
controlnet
cuda-cpu

huchenlei:
I will merge this code as unittests on IPAdapter is added in https://github.com/Mikubill/sd-webui-controlnet/pull/2252

Issue #1432, opened 2023-05-26
[Feature]: Do not return ControlNet images via API


andypotato:
### Is there an existing issue for this?

- [X] I have searched the existing issues and checked the recent builds/commits of both this extension and the webui

### What happened?

When creating images via the API using ControlNet, the control images will be sent back as part of the result images array. While this may be useful in some cases, most of the time it is just unnecessary data traffic.

Therefore I suggest to have an option to remove the control images from the resulting images array, similar to how the A1111 API already has a "send_images" parameter.

### Steps to reproduce the problem

Create any txt2img or img2img API request with a ControlNet module enabled.

### What should have happened?

See feature request description - returning the original control images via the API should be optional.

### Commit where the problem happens

Feature request

### What browsers do you use to access the UI ?

Google Chrome

### Command Line Arguments

```Shell
--api
```

### List of enabled extensions

ControlNet

### Console logs

```Shell
No logs necessary
```

MagicReader:
I want it too

Dadacticiel:
Hello ! As a workaround, i added those lines in stable diffusion webui code to remove controlnet images in api response.
File **stable-diffusion-webui\modules\api\api.py**, around line 448, add the try/except segment :
```     
        if not img2imgreq.include_init_images:
            img2imgreq.init_images = None
            img2imgreq.mask = None
            try:
                for index, controlnet_arg in enumerate(img2imgreq.alwayson_scripts["ControlNet"]["args"]):
                    if "image" in controlnet_arg and controlnet_arg["image"] is not None and "image" in controlnet_arg["image"]:
                        img2imgreq.alwayson_scripts["ControlNet"]["args"][index]["image"]["image"] = None
            except AttributeError:
               pass

**Comment by wasa4587** [comment]
here's an example
```
{
...,
       "alwayson_scripts": {
         "controlnet": {
           "args": [
             {
               "input_image": openposeBase64,
               "model": "control_v11p_sd15_openpose [cab727d4]",
               save_detected_map: true,
               "pixel_perfect": false
             },
             {
                "input_image": depthBase64,
                "model": "control_v11f1p_sd15_depth [cfd03158]",
                save_detected_map: true,
                "pixel_perfect": false
            } 
           ]
         }
       }
}
```

PR #2251, opened 2023-11-11
:sparkles: Add new API option send_detected_map

huchenlei:
Closes https://github.com/Mikubill/sd-webui-controlnet/issues/1432.

This PR adds a new API option `send_detected_map`. This option allows users prevent the detected map image in the response image array to save web traffic.

Notes: 
- The option will be captured in the ControlNet infotext, but will not be adjustable in WebUI for now.
- The option is per ControlNet unit, so if you use multiple ControlNet units, you need to specify this option in every unit.
- Adding a global level args in the API is not possible as it would need change in A1111 repo.

### DIFF
--- a/internal_controlnet/external_code.py
+++ b/internal_controlnet/external_code.py
@@ -1,3 +1,4 @@
+from dataclasses import dataclass
 from enum import Enum
 from copy import copy
 from typing import List, Any, Optional, Union, Tuple, Dict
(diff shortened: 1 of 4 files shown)


wasa4587:
correction the parameter name is `save_detected_map`

```
{
...,
       "alwayson_scripts": {
         "controlnet": {
           "args": [
             {
               "input_image": openposeBase64,
               "model": "control_v11p_sd15_openpose [cab727d4]",
               save_detected_map: true,
               "pixel_perfect": false
             },
             {
                "input_image": depthBase64,
                "model": "control_v11f1p_sd15_depth [cfd03158]",
                save_detected_map: true,
                "pixel_perfect": false
            } 
           ]
         }
       }
}
```

Issue #2222, opened 2023-11-04
[DevTask] Automatically filter ControlNet models based on active SD model version


huchenlei:
There are many issues caused by users accidentally choosed wrong ControlNet model that does not fit the active SD model. For example, using SDXL ControlNet model for SD1.5 SD model. #2221, #2155

We should implement a mechanism to automatically filter ControlNet model selections based on current active SD model in A1111, and give meaningful warning/error messages if mismatches are detected in A1111 API calls.

sdbds:
I remember that sd-webui 1.6 had the ability to filter the corresponding version of LORA. Maybe refer to that section.

PR #2268, opened 2023-11-20
🔊 Meaningful error message for sd version conflict

huchenlei:
Closes https://github.com/Mikubill/sd-webui-controlnet/issues/2222.

This PR adds a check on sd version conflict between ControlNet model and sd model. This can be useful for debugging purpose for API callers, as they do not have the auto model filtering in the UI.

Sample error message:
```shell
Exception: ControlNet model kohya_controllllite_xl_canny [2ed264be](StableDiffusionVersion.SDXL) is not compatible with sd model(StableDiffusionVersion.SD1x)
```

### DIFF
--- a/scripts/controlnet.py
+++ b/scripts/controlnet.py
@@ -20,7 +20,7 @@ from scripts.controlnet_lllite import PlugableControlLLLite, clear_all_lllite
 from scripts.controlmodel_ipadapter import PlugableIPAdapter, clear_all_ip_adapter
 from scripts.utils import load_state_dict, get_unique_axis0
 from scripts.hook import ControlParams, UnetHook, HackedImageRNG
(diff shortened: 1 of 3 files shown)


foglerek:
Hello, this appears to have created a bug when using Reference (`Unable to determine version for ControlNet model 'None'.`) — from my testing, it seemed that the reference image was not getting applied after this update, and the above error message was displayed. I believe it's due to this change:

```python
        cnet_sd_version = StableDiffusionVersion.detect_from_model_name(unit.model)

        if cnet_sd_version == StableDiffusionVersion.UNKNOWN:
            logger.warn(f"Unable to determine version for ControlNet model '{unit.model}'.")
            return
```

huchenlei:
Thanks for pointing that out. Fixing rn.

PR #2275, opened 2023-11-21
:memo: Update to version v1.1.419

huchenlei:


trklmsr:
Doesn't work with Automatic1111.
Preprocessor and controlnet models do not work.
I get results as if controlnet is not installed at all.

huchenlei:
> Doesn't work with Automatic1111. Preprocessor and controlnet models do not work. I get results as if controlnet is not installed at all.

please file an issue with reproduction steps. Thanks.

Issue #2287, opened 2023-12-01
not an issue - question on densepose 


johndpope:
is there a reason why there's no densepose for controlnet? seems superior option for body detection.

https://www.reddit.com/r/StableDiffusion/comments/17yirws/openpose_alternatives_humansd_densepose/

https://github.com/facebookresearch/detectron2/issues/5166

related
https://github.com/lllyasviel/ControlNet/issues/445

sdbds:
![image](https://github.com/Mikubill/sd-webui-controlnet/assets/8085926/06310fd6-1c23-408b-b603-e0a772101505)
![image](https://github.com/Mikubill/sd-webui-controlnet/assets/8085926/b82e52e6-6f2d-46c0-ba6e-fef51af59dbb)

Actually densepose should be categorized as a semantic segmentation. 
But since it doesn't have specific semantics, it doesn't work very well.
 I had a plan for a fashion segmentation, but I was too busy to deploy it.

johndpope:
looks mint - is this done with public code - or stuff you haven't released? It's exactly what I want to do.
this guy has this model - but I don't understand how it can work. 
https://civitai.com/models/120149/controlnet-for-densepose

I did look through detectron2 stuff in repo - but there's no specific densepose references in code unless I'm mistaken.

UPDATE - I did check again and saw - but it doesn't correspond specifically to that code from facebook.
<img width="1036" alt="Screenshot 2023-12-04 at 3 33 41 pm" src="https://github.com/Mikubill/sd-webui-controlnet/assets/289994/dfef1040-4d0d-40c9-b11f-80c17dcd4786">



the segmentation - yes did see - but again does not equal the densepose.
do these lines / squiggles not carry enough information / annotations - could we extend it to give us this ? 
<img width="516" alt="Screenshot 2023-12-04 at 3 36 12 pm" src="https://github.com/Mikubill/sd-webui-controlnet/assets/289994/cf06c8a8-fc0c-44ff-a97a-e8bc0daf428d">

I'm very keen to help and see this live. would be delighted to test a branch of fashion stuff. 

I'm using chatgpt to decipher more complex code and if you haven't tried it - I recommend.
You can ask it to code up complicated algorithms and it will spit it out for you. 
It's read your repo and knows it inside out. Maybe it can help you get the fashion stuff over the line.


here's some sample code I threw at it


**comment this code from sd-webui-controlnet/controlnet-lora.py import torch**

```

[...]

PR #2314, opened 2023-12-10
Add Densepose (TorchScript)

Fannovel16:
Request from https://github.com/Mikubill/sd-webui-controlnet/issues/2287
Compiled to TorchScript through tracing and scripting from [R_50_FPN_DL_s1x](https://github.com/facebookresearch/detectron2/blob/main/projects/DensePose/doc/DENSEPOSE_IUV.md#-improved-baselines-deeplabv3-head) as Detectron2 is a pain to install to end users
Reference: https://github.com/Fannovel16/comfyui_controlnet_aux/blob/main/src/controlnet_aux/densepose/__init__.py
Expected result:

Compatiable ControlNets: Implementation on [CivitAI](https://civitai.com/models/120149/controlnet-for-densepose) (parula color map), from [MagicAnimate](https://huggingface.co/zcxu-eric/MagicAnimate/blob/main/densepose_controlnet/diffusion_pytorch_model.safetensors) (viridis color map)


lllyasviel:
someone having free time can have a test @huchenlei @continue-revolution @sdbds

sdbds:
> someone having free time can have a test @huchenlei @continue-revolution @sdbds

I will test it.

sdbds:
image
image

Fix some bugs and rename def, it works well now.

image

I add densepose to OpenPose filters,too.
Althrough it should be segments...

sdbds:
Looks like there are no other issues, waiting for others to test results and review.

Fannovel16:
There are two possible issues I can think of:
1. It may give weird results on some aspect ratios as I used `torch.jit.trace`
2. Does it work well on non-cuda devices like MPS?

Kinglord:
I'm having lots of issues with poor results using this, ranging from artifacts, to suddenly missing body segments and large noise spikes that seem random from frame to frame. If I use this to process video frames I would say at least 60% of the processed frames contain issues (I don't think this is unique, it appears in much of the samples on Magic Animate). I just have to think there's a better method out there even if it's a PITA to install/deal with? Does the Detectron2 implementation suffer from the same issues? I've tried a bunch of different videos/images as source drivers, so unless I need to format my images in a certain way to fix this I'm not sure how else it could be improved. It feels like to me this is useful as a "helper" cnet, but I would never use it alone or at high strength due to these errors.

sdbds:
> I'm having lots of issues with poor results using this, ranging from artifacts, to suddenly missing body segments and large noise spikes that seem random from frame to frame. (You can see in the example @sdbds provided it has errors on the chest/shoulders and has the wrong angle of the body) If I use this to process video frames I would say at least 60% of the processed frames contain issues (I don't think this is unique, it appears in much of the samples on Magic Animate). I just have to think there's a better method out there even if it's a PITA to install/deal with? Does the Detectron2 implementation suffer from the same issues? I've tried a bunch of different videos/images as source drivers, so unless I need to format my images in a certain way to fix this I'm not sure how else it could be improved. It feels like to me this is useful as a "helper" cnet, but I would never use it alone or at high strength due to these errors.

I think it is original repo problem.
@Fannovel16 Did u try to remove background like anime_face_segment?
Densepose is another way about segement,maybe it will get better result without background.

Kinglord:
I did notice a few things that I think are "well known" densepose factors but I was new to the technology. Basically full body input is ideal but more the merrier - you do NOT want to crop before you preprocess. I did try and play with background removal and it didn't seem to help in any noticeable way, sometimes it made it worse (guessing the model is trained with backgrounds) but I didn't try a large sample set there. I think it might also be tied to the format/size of the driver data but I haven't played with it enough to know for sure. My hunch ATM is that full body, 512x512 driving data is the ideal any the farther you veer from that the worse results you will get. That said I still see some horrible results with data that doesn't strike me as something that would be that hard. I can tell that "human" proportions also seems to matter a lot. For example I have a very simple image of a woman looking away from the camera in a full body shot. The preprocessor mutilates her legs, when I tried to scrunch the video down to other sizes it gets worse, and at 512x512 it loses her upper torso entirely.

ComfyUI_temp_gkxya_00002_
ComfyUI_temp_gkxya_00003_

johndpope:
nice work @Mikubill - 

Could it be  "trivial" to swap out the densepose to use dwpose / animal pose torchscript?
I see comfyui has achieved this in last 2 weeks.

https://github.com/Fannovel16/comfyui_controlnet_aux/commit/4da9ae776c3fb0e205622f87afc790ef373964f1

dajes:
I don't know if someone is interested in this but I modified the original DensePose code to make it compilable and provided the compiled models [here](https://github.com/dajes/DensePose-TorchScript). You only need torch, torchvision and opencv to run this compiled model

sdbds:
@huchenlei I think it can be merged now.Any more suggestion?

sdbds (approved):

EtienneDosSantos:
> I add densepose to OpenPose filters,too.
> Althrough it should be segments...

Why are the densepose preprocessors in OpenPose filters while they're not in those of segmentation? In my tests it didn't even work together with the OpenPose model (SD 1.5).

sdbds:
> > I add densepose to OpenPose filters,too.
> > Althrough it should be segments...
> 
> Why are the densepose preprocessors in OpenPose filters, while they're not in those of segmentation? In my tests it didn't even work together with the OpenPose model (SD 1.5).

it use another model in civitai.
https://civitai.com/models/120149/controlnet-for-densepose

continue-revolution:
I tend to separate DensePose into another category.

EtienneDosSantos:
> > > I add densepose to OpenPose filters,too.
> > > Althrough it should be segments...
> > 
> > 
> > Why are the densepose preprocessors in OpenPose filters, while they're not in those of segmentation? In my tests it didn't even work together with the OpenPose model (SD 1.5).
> 
> it use another model in civitai. https://civitai.com/models/120149/controlnet-for-densepose

I see, wasn't aware of that, thx for pointing that out!

PR #2356, opened 2023-12-25
fix image invert detect

liubo0902:
fix detect api mode, invert cannot return image base64


huchenlei:
Thanks for spotting the problem and the fix!

PR #2359, opened 2023-12-26
support clip model cache

liubo0902:
Since clip models are bigger than 1G, unloading and loading the clip model every call costs a lot time for model loading.


huchenlei:
I believe we already have clip cache here:

https://github.com/Mikubill/sd-webui-controlnet/blob/b25fe459139a0cf0ab0934a5907c8371813ca338/scripts/processor.py#L348-L356

I am going to close this PR now. Please re-open if you can prove that the cache here is really necessary.

PR #2366, opened 2023-12-28
:art: Hide loopback checkbox in non-batch config

huchenlei:
Loopback option has no effect in non-batch generation. Hide the checkbox in non-batch tab so that it does not confuse users.


ljleb:
This is inaccurate. Loopback can be used on a unit that is in image mode, if a different unit is in batch mode.

huchenlei:
@ljleb I have fixed the logic. Now loopback checkbox will show if any of the unit has batch enabled.

Issue #2249, opened 2023-11-11
[Request]: User defined scheduling


kft334:
### Is there an existing issue for this?

- [X] I have searched the existing issues and checked the recent builds/commits of both this extension and the webui

### What happened?

Please add user defined scheduling, separately for firstpass/highrespass, at least to the API. The current weight, start, end for all passes is too restrictive and inflexible in some cases. In the API it can be added as an optional argument that overrides weight, start and end if set. It would not affect existing API users since it will be treated as optional. If only the first pass schedule is provided, it would apply to both passes. If both schedules are provided then they would be applied to their respective passes. 

Example (lines split for readability):
0.2@0.1,
0.5@0.3,
0.2@0.5,
0.1@0.7

If this string is provided it will override weight, start and end. In this case start is 0.1 and the strength starts at 0.2 and scales up linearly to 0.5 by 30% and so on until finally reaching the end at 0.7. Strength is then 0 until 100%. In this case only first pass is provided so it would also apply to the second pass if second pass is enabled.

In order for it not to be awkward to work with in the webui, add graph widgets so that users don't even have to work with the string. I can provide a working example of this since I have it implemented in my SD app for another extension (loractl) and it works extremely well and is user friendly. I realize that you're trying to keep ControlNet as user friendly as possible but it is arguably the most powerful extension in webui so it should be as flexible and limitless as possible. If implemented correctly though, it will not add much complexity to the user experience but will add another layer of flexibility.

### Steps to reproduce the problem

na

### What should have happened?

na

### Commit where the problem happens

na

### Command Line Arguments

```Shell
na
```

### List of enabled extensions

na

### Console logs

```Shell
na
```

huchenlei:
I believe the `advanced_weight` is what you are requesting. Currently the field is not wired though. I think we just need to figure out a good semantics for the API to specify `advanced_weight`.

https://github.com/Mikubill/sd-webui-controlnet/blob/e2e6a1b4941a8386c388280ffe7c4304e4981c6d/scripts/hook.py#L594-L595

kft334:
Does that support second pass too atm? For the API how about just using the weight param? If weight is passed as a number then it's simple weight/start/end. If it's passed as a string process the string and apply control_scales. Won't affect existing users using numbers and no extra params needed.

weight: "0.2@0.0,0.6@0.5,1.2@1.0;1.0@0.0,0.5@0.5,0.5@1.0"
start: doesn't matter if weight is string
[...]

PR #2368, opened 2023-12-30
:sparkles: Support advanced_weighting in API

huchenlei:
Relates to discussion in #2249.

This PR allows user to specify weight for each ControlNet layer via API.


huchenlei:
The default soft injection in current ControlNet code uses
```python
advanced_weighting = [0.85 ** float(12 - i) for i in range(13)]
```

Here I tried replacing the 0.85 constant with some other values via API in the example code snippets:
0.3
txt2img-0fac0 3
0.5
txt2img-0fac0 5
0.8
txt2img-0fac0 8
canny map
txt2img-1fac0 3

I am not very convinced that this should be exposed in A1111 UI, so this feature is implemented as API-only.

huchenlei:
Tested more values:
0.75
txt2img-0fac0 75
0.8
txt2img-0fac0 8
0.95
txt2img-0fac0 95

Issue #2369, opened 2023-12-30
[Feature Request] Crop input image checkbox for each ControlNet unit in inpaint mask only


huchenlei:
Original Request: https://github.com/Mikubill/sd-webui-controlnet/pull/2365#issuecomment-1872310608

Let user decide whether the ControlNet input image should be cropped according to A1111 mask when using A1111 inpaint mask only. Currently, the setting is global to all ControlNet units.

PR #2377, opened 2023-12-31
:sparkles: Control inpaint crop per ControlNet unit

huchenlei:
Closes https://github.com/Mikubill/sd-webui-controlnet/issues/2369.

This PR adds a new checkbox to control whether the ControlNet input image should be cropped to A1111 mask or not.

The checkbox is only visible if the user
- in img2img inpaint tab
- have inpaint area set to masked only
- check upload independent control image


huchenlei:
Note:
Original behaviour was cropping input image. Now you need to check that checkbox to restore the original behaviour.

PR #2379, opened 2024-01-01
:sparkles: Add hr_option (Both, High res only, Low res only)

### LINKED ISSUE #2378: [Feature Request] More hires fix options
Original discussion: #1826.

We should provide option per ControlNet unit to let user pick how the ControlNet unit should apply:
- To both pass
- Only low res pass
- Only high res pass

huchenlei:
Closes #2378.

This PR removes global option `control_net_no_high_res_fix`, and adds the radio selection for hr option per ControlNet unit.

### DIFF
--- a/internal_controlnet/external_code.py
+++ b/internal_controlnet/external_code.py
@@ -7,6 +7,7 @@ from modules import scripts, processing, shared
 from scripts import global_state
 from scripts.processor import preprocessor_sliders_config, model_free_preprocessors
 from scripts.logging import logger
+from scripts.enums import HiResFixOption
 
 from modules.api import api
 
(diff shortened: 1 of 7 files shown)


huchenlei:
Previously disabling highres pass was available so I guess there is sufficient usecases being explored. Here just some novel uses enabled by this PR.

Now you can use tile model to control highres pass.

ControlNet input image:
download (2)

Step1: non-highres generation, with canny edge unit:


Step2: Reuse seed. Highres generation with 2 units:
- canny edge (set to lowres pass only)
- tile (input is the generation result from step1) (set to highres pass only)
00009-1924817058_hires

I am not sure whether this method is superior than the img2img upscale method (tile + SD Ultimate upscale script), but at least there is another option now 😄.

huchenlei:
Tile + SD Ultimate upscale script with same seed:
00001-1924817058_img2img

The highres fix method seem to generate less noisy irrelevant details.

huchenlei:
image

huchenlei:
This feature is also requested in https://github.com/Mikubill/sd-webui-controlnet/issues/2249.

huchenlei:
Direct t2i with tile model, 2x size (1024x1536) and same seed:

00000-1924817058_t2i

Hires result has some slight difference in details, probably because added noise during diffusion process. Conclusion is probably not worth doing tile with hires fix?

dark-edgelord:
I don't see this option at all in the UI anywhere, even with hiresfix enabled.

huchenlei:
You need A1111 1.7.0+ for the UI to work properly.

dark-edgelord:
I have that, but I found out what the issue is. It only shows up when clicking the hires fix check box, not the accordion (that also ticks the box automatically).

image

Clicking green works, red doesn't.

PR #2415, opened 2024-01-05
:wrench: Auto disable photopea edit on confirm alert cancel

huchenlei:
As requested by @w-e-w, I made disabling the feature easier. Now click cancel will automatically check the checkbox in settings tab, and hide all edit buttons in current session.


w-e-w:
LGTM

w-e-w:
just have another thought
this approach you're using now doesn't have good multi-user compatibility
basically if there's a bunch of users and one user decides to click cancle, it will be disabled for all users ....

huchenlei:
Do we have anything in the settings that is per-user basis now?

w-e-w:
let me try and came up with a different solution see if you like it
not per user but more flexible than this

PR #2417, opened 2024-01-05
alternative photopea popup warning

w-e-w:
@huchenlei see how you like this approach, feel free to make modifications

---

## alternative photopea popup warning

add a new setting `Photopea popup warning` this controls if you pop up is shown before reuning photopea, default enabled

provides a middle ground for users that may wish to use photpea form time to time doesn't want to accidentally use it

>sadly no multi-user diffrent config support

## behavior
when `Photopea popup warning` is enable
pop up will be shown on the first time that a user tries to use photopea

if the user choose cancel nothing happens..... well it cancels and nothing else happes
if the user clicks on the photopea again the cycle repeats

if the user choose OK photopea is loaded
the value is set so that the pop-up doesn't show up again for for THIS browser session (until the page reloads)

---

if the user disables `Photopea popup warning` then photopea is loaded immediately agter clicking the icon

---

if `Photopea integration` is disabled, .... Photopea doesn't exist and none of the above will happen

https://github.com/Mikubill/sd-webui-controlnet/assets/40751091/99993c1a-f88d-4a97-9ef4-ae29e1ee03cb

### DIFF
--- a/javascript/photopea.js
+++ b/javascript/photopea.js
@@ -375,32 +375,19 @@
     }
   }
 
-  function firstTimeUserPrompt() {
-    if (localStorage.getItem("ControlNetPhotopeaEditPrompted")) {
-      return true;
-    } else {
-      localStorage.setItem("ControlNetPhotopeaEditPrompted", "true");
-    }
-    const promptMsg = "This is the first time you use photopea edit feature. The preprocess results are " +
-      "going to be send to https://photopea.com for edit.\n" +
-      "- Click OK: proceed.\n" +
-      "- Click Cancel: abort and disable photopea edit feature.";
-    const confirmed = confirm(promptMsg);
+  let photopeaPopupPopupShown = false;
 
-    if (!confirmed) {
-      // Hide all edit buttons in current session.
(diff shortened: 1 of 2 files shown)


javascript/photopea.js
```diff
@@ -375,32 +375,19 @@
     }
   }
 
+  let photopeaPopupPopupShown = false;
```
huchenlei, line 378: Can we rename these variables?
- `photopeaPopupPopupShown` => `photopeaWarningShown`
- `controlnet_disable_photopea_popup` => `controlnet_photopea_warning`

huchenlei (changes requested):
Overall looks good. Just one nit of distinguishing the modal popup and warning popup in variable naming.

javascript/photopea.js
w-e-w: sure, I didn't put too much thought in the naming of those variables
I did say feel free to make modifications, 
in case you're not aware you are actually able to push to my Fork because `Allow edits and access to secrets by maintainers` es checked
image

also maybe some of my wording of the messages are not as good
if you want to change those go ahead

huchenlei (approved):

PR #2461, opened 2024-01-12
Fix editor window size with SD.Next

lbeltrame:
Default themes in SD.Next enforce a maximum width for the column where the editor window opens, thus constraining it to a very small size.

Override `max-width` so that it is not affected by it. Fixes vladmandic/automatic#2594.


huchenlei:
LGTM

Issue #2451, opened 2024-01-11
[Enhancement] Support region condition on each ControlNet unit


huchenlei:
https://github.com/cubiq/ComfyUI_IPAdapter_plus has support for attn_mask on IP-Adapter to limit the prompt effect to a specific region in the generated image. It is similar to regional prompter, but for IP-Adapter embeddings.

In ControlNet, some models (canny, openpose) has pure black being empty space, i.e. no control will be applied to pure black regions. However, for other models (tile, depth), there is no way to specify the control region. Pure black region has its own semantics for these models.

We should build an integrated editor that allow easy edit of control regions, and use the mask from the editor to limit the region of control.

huchenlei:
TODOs:
- [ ] Currently mask is only used by inpaint preprocessors. Need to find a way to make masking universal.
- [ ] Make sure masking works for HiRes. Fix.
- [ ] Support masking for ControlNet models
- [ ] Support masking for IP-Adapters

mikedreambooth:
This is something I have thought about for a while but I don't know enough about the extension to contribute here! I had noticed if you use a higher weight on a canny control for example, certain models will generate images with little detail outside of the canny regions. So it's a trade of at the minute.

An option similar to the inpaint tab where you can choose to fill the masked region with original/latent noise/latent nothing could be useful here too?

huchenlei:
Feature implemented in Issue solved in https://github.com/lllyasviel/stable-diffusion-webui-forge

Issue #1768, opened 2023-07-05
img2img+inpaint broken


lllyasviel:
someone told me that a1111 img2img + controlnet inpaint_global_harmonious is not working. I tested, it seems broken. 

To reproduce:

1. go to a1111 img2img, select img2img (the simple img2img without inpaint, not to confuse with a1111 img2img inpaint)
2. go to controlnet, select inpaint_global_harmonious
3. cannot upload image, cannot draw inpaint scribbles

This also make other modes broken, like a1111 img2img sketch + inpaint_global_harmonious, a1111 img2img inpaint sketch + inpaint_global_harmonious, a1111 img2img sketch + inpaint_only, a1111 img2img inpaint sketch + inpaint_only, etc

Besides, it seems that inpaint_global_harmonious's functionality to make soft transition is also broken in a1111 img2imh inpaint. Previously users can draw masks with larger holes in a1111 and draw different masks with a bit smaller holes in inpaint_global_harmonious to reduce seams in inpaint

huchenlei:
https://github.com/Mikubill/sd-webui-controlnet/pull/1763 Disallows use of ControlNet input in img2img inpaint. If you click to upload an image it will display an alert let user use A1111 inpaint input.

If global harmonious requires the ControlNet input inpaint, for now, user can select `All` control type and select preprocessor/model to fallback to previous behaviour.

Or you can revert https://github.com/Mikubill/sd-webui-controlnet/pull/1763 for now. I will reland it later with global harmonious inpaint issue fixed.

huchenlei:
If the revert is not straightforward, you can remove calls to https://github.com/Mikubill/sd-webui-controlnet/blob/f2a541dd8fc10b34c415b14412a97f596796306c/javascript/active_units.js#L142, which should remove the restrictions on inpaint.

huchenlei:
Old behaviour restored in https://github.com/Mikubill/sd-webui-controlnet/pull/1769.

huchenlei:
> 

Thanks for finding that comment! We should really document it somewhere more accessible. The new user would never be able to figure out how to properly use both masks together without guidance.

PR #2468, opened 2024-01-14
:hammer: Refactor choose_input_image logic

huchenlei:
This PR is the prework of a larger effort in https://github.com/Mikubill/sd-webui-controlnet/issues/2451.

The PR consolidates the logic of choosing `resize_mode` into `Script.choose_input_image`. In summary, `resize_mode` goes along with the input image source. 
- If input image is from A1111 img2img input, A1111 resize mode is used.
- If input image is from ControlNet unit, ControlNet unit's resize mode is used.

Extra full coverage unittests are added to ensure that behaviours of various inpaint setups are preserved. Those unittests are skipped on CI runs.

Command to run those unittests:
`python -m pytest -vv --junitxml=test/results.xml --cov .\extensions\sd-webui-controlnet\ --cov-report=xml --verify-base-url .\extensions\sd-webui-controlnet\tests\web_api\full_coverage\inpaint_test.py`

You need to set environment variable `CONTROLNET_TEST_FULL_COVERAGE` to any value to enable full coverage tests.
Run first with `is_set_expectation_run` set to `True` to establish a baseline.

**Breaking behaviour**:
Before you can set both A1111 img2img input image and ControlNet unit input image, and set mask on A1111 input. Now the behaviour no longer works, as we expect the mask to be on ControlNet unit if ControlNet input image is present.

I do not see any point of doing this use case, as if you want to use A1111 mask, you would not upload an independent controlnet input image. So I guess it would be OK to ban this use case. In the updated code, this case will raise an exception saying that no mask found on ControlNet input image.


huchenlei:
Note: all use cases mentioned in https://github.com/Mikubill/sd-webui-controlnet/issues/1768 are still functional as before.

Issue #2436, opened 2024-01-08
[Enhancement] Run CLIP (IP-Adapter preprocessor) on CUDA


huchenlei:
Currently all CLIP preprocessors run on CPU. See
https://github.com/Mikubill/sd-webui-controlnet/blob/bb9483d46f5a932fd35e8b4d04a3fdcc02dd9ff1/annotator/clipvision/__init__.py#L112-L132

PR #2474, opened 2024-01-15
feat: significantly optimize the time consumption of clip vision

storyicon:
In the inference service using ControlNet, I observed a significant increase in time consumption when utilizing clip vision-related preprocessors (e.g. `adapter_clip_sd15`). The time fluctuates between 8s and 250s, depending on machine load. The typical time is around 10s. My machine is equipped with an A10 GPU, 96C CPU, and 251GB memory.

Upon investigation, I found a substantial portion of the time is spent in the following section: https://github.com/Mikubill/sd-webui-controlnet/blob/main/annotator/clipvision/init.py#L101.

Further exploration revealed that these time-consuming operations are primarily due to extensive initialization, such as:
https://github.com/huggingface/transformers/blob/main/src/transformers/models/clip/modeling_clip.py#L333-L334
https://github.com/huggingface/transformers/blob/main/src/transformers/models/clip/modeling_clip.py#L241-L244
This includes operations like nn.Linear, among others.

This issue is related to https://github.com/huggingface/transformers/issues/21913.

After implementing this optimization, the average time consumption for `clip vision` can be compressed to at least 1/6 of the previous time. The results are highly stable, and there is no inconsistency in the inference outcomes.

In my case, before optimization, even on a completely idle machine, the preprocessing time for `clip vision` took at least 8 seconds. Now, it only requires 1.23 seconds.

### DIFF
--- a/annotator/clipvision/__init__.py
+++ b/annotator/clipvision/__init__.py
@@ -98,23 +98,22 @@ class ClipVisionDetector:
         if not os.path.exists(file_path):
             load_file_from_url(url=self.download_link, model_dir=self.model_path, file_name=self.file_name)
         config = CLIPVisionConfig(**self.config)
-        self.model = CLIPVisionModelWithProjection(config)
-        self.processor = CLIPImageProcessor(crop_size=224,
-                                            do_center_crop=True,
-                                            do_convert_rgb=True,
-                                            do_normalize=True,
-                                            do_resize=True,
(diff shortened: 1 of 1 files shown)


huchenlei:
I think this closes https://github.com/Mikubill/sd-webui-controlnet/issues/2436. Thanks for the fix!

storyicon:
cool, so efficient.

Issue #2482, opened 2024-01-16
[Bug]: Tmp Folder [WinError5] Permission Error


Talshiar2:
### Is there an existing issue for this?

- [X] I have searched the existing issues and checked the recent builds/commits of both this extension and the webui

### What happened?

I tried installing ControlNet via link and via extension tab, always run in this problem, other extensions loaded problem free.

### Steps to reproduce the problem

1. Go to extension,  for [sd-webui-controlnet](https://github.com/Mikubill/sd-webui-controlnet.git)
manipulations]
2. Press/hit "install"

### What should have happened?

It should have installed ControlNet

### Commit where the problem happens

webui: 1.7.0
controlnet: n/a --> happens at installing it

### What browsers do you use to access the UI ?

Google Chrome

### Command Line Arguments

```Shell
set COMMANDLINE_ARGS=--use-directml --onnx --skip-torch-cuda-test
```

### List of enabled extensions

canvas-zoom
sd-webui-animatediff
sd-webui-aspectratio-helper

### Console logs

```Shell
To create a public link, set `share=True` in `launch()`.
Startup time: 12.8s (prepare environment: 4.6s, import torch: 2.6s, import gradio: 1.5s, setup paths: 1.6s, initialize shared: 1.1s, other imports: 0.3s, load scripts: 1.7s, create ui: 0.8s, gradio launch: 0.2s).
*** Error completing request
*** Arguments: ('https://github.com/Mikubill/sd-webui-controlnet.git', ['ads', 'localization', 'installed'], 0, '') {}
    Traceback (most recent call last):
      File "F:\andreai\stable-diffusion-webui-directml\modules\call_queue.py", line 57, in f
        res = list(func(*args, **kwargs))
      File "F:\andreai\stable-diffusion-webui-directml\modules\ui_extensions.py", line 399, in install_extension_from_index
        ext_table, message = install_extension_from_url(None, url)
      File "F:\andreai\stable-diffusion-webui-directml\modules\ui_extensions.py", line 387, in install_extension_from_url
        raise err
      File "F:\andreai\stable-diffusion-webui-directml\modules\ui_extensions.py", line 379, in install_extension_from_url
        os.rename(tmpdir, target_dir)
[...]

PR #2493, opened 2024-01-18
Use pip for HandRefinerPortable

akx:
After https://github.com/huchenlei/HandRefinerPortable/pull/2 is merged (making that repository Pip installable), this PR changes sd-webui-controlnet to not use a Git submodule for it anymore.

This was discussed with @w-e-w and @huchenlei on the A1111 discord.

This will not clean up any previously cloned submodule (nor deinit submodules), though.

Because we have removed submodule `hand_refiner_portable`, there should no longer be any issue of renaming dir. We suspect previously it was `git submodule update` holding some resources, which cause `os.rename` to fail with PermissionError.
- Closes https://github.com/Mikubill/sd-webui-controlnet/issues/2482
- Closes https://github.com/Mikubill/sd-webui-controlnet/discussions/2488

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,0 @@
-[submodule "annotator/hand_refiner_portable"]
-	path = annotator/hand_refiner_portable
-	url = https://github.com/huchenlei/HandRefinerPortable
--- a/install.py
+++ b/install.py
@@ -1,7 +1,5 @@
 import launch
-import git  # git is part of A1111 dependency.
 import pkg_resources
-import os
 import sys
 import platform
 import requests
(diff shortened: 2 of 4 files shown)


huchenlei (approved):
LGTM. Thanks!

akx:
@huchenlei To be clear, this will _not_ deinitialize the submodule (if already initialized) nor delete the folder, but they won't do anything. Users _can_ clean up with `git submodule deinit --all` and removing the `hand_refiner_portable` submodule folder, but it shouldn't be necessary.

In any case, glad to help!

Issue #2498, opened 2024-01-19
[Bug]: Warning: Failed to install handrefinerportable@git+https://github.com/huchenlei/HandRefinerPortable.git, some preprocessors may not work.


swumagic:
### Is there an existing issue for this?

- [X] I have searched the existing issues and checked the recent builds/commits of both this extension and the webui

### What happened?

Installing sd-webui-controlnet requirement: handrefinerportable@git+https://github.com/huchenlei/HandRefinerPortable.git
Couldn't install sd-webui-controlnet requirement: handrefinerportable@git+https://github.com/huchenlei/HandRefinerPortable.git.
Command: "J:\StableDiffusion\sdwebui\py310\python.exe" -m pip install handrefinerportable@git+https://github.com/huchenlei/HandRefinerPortable.git --prefer-binary
Error code: 2
stdout: Looking in indexes: https://pypi.org/simple, https://download.pytorch.org/whl
Collecting handrefinerportable@ git+https://github.com/huchenlei/HandRefinerPortable.git
  Cloning https://github.com/huchenlei/HandRefinerPortable.git to c:\users\admin\appdata\local\temp\pip-install-5vu9tjyg\handrefinerportable_4aae1388755947fd8fce07cb91d99ad7
  Resolved https://github.com/huchenlei/HandRefinerPortable.git to commit d9c5506a1c13b396499dfdbaada9b7650e5d6d65
  Installing build dependencies: started
  Installing build dependencies: finished with status 'done'
  Getting requirements to build wheel: started
  Getting requirements to build wheel: finished with status 'done'

stderr:   Running command git clone --filter=blob:none --quiet https://github.com/huchenlei/HandRefinerPortable.git 'C:\Users\admin\AppData\Local\Temp\pip-install-5vu9tjyg\handrefinerportable_4aae1388755947fd8fce07cb91d99ad7'
ERROR: Exception:
Traceback (most recent call last):
  File "J:\StableDiffusion\sdwebui\py310\lib\site-packages\pip\_internal\cli\base_command.py", line 180, in exc_logging_wrapper
    status = run_func(*args)
  File "J:\StableDiffusion\sdwebui\py310\lib\site-packages\pip\_internal\cli\req_command.py", line 245, in wrapper
    return func(self, options, args)
  File "J:\StableDiffusion\sdwebui\py310\lib\site-packages\pip\_internal\commands\install.py", line 377, in run
    requirement_set = resolver.resolve(
[...]

PR #2499, opened 2024-01-19
Update requirements.txt

swumagic:
Warning: Failed to install handrefinerportable@git+https://github.com/huchenlei/HandRefinerPortable.git, some preprocessors may not work.
ModuleNotFoundError: No module named 'hatchling'
#2498


huchenlei:
Issue fixed in #2504. Now we directly install from a prebuild wheel.

PR #2500, opened 2024-01-19
Devices xpu and mps

kingljl:
Obtain the current system compatible drivers through the built-in devices, support XPU or MPS, and speed up calculations.

The place where the model was loaded was not changed to avoid problems with the load model, only the annotator was changed

code:
```python
torch.device('cuda' if torch.cuda.is_available() else 'cpu'))  replace  torch.device(devices.get_optimal_device_name()))

def get_optimal_device_name():
    if torch.cuda.is_available():
        return get_cuda_device_string()
    if has_mps():
        return "mps"

    if has_xpu():
        return xpu_specific.get_xpu_device_string()
    return "cpu"
```

### DIFF
--- a/annotator/clipvision/__init__.py
+++ b/annotator/clipvision/__init__.py
@@ -79,10 +79,11 @@ downloads = {
 
 
 clip_vision_h_uc = os.path.join(os.path.dirname(os.path.abspath(__file__)), 'clip_vision_h_uc.data')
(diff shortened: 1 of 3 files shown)


huchenlei:
We would prefer calling `devices.get_device_for("controlnet")` instead of directly calling `get_optimal_device_name`.

kingljl:
> We would prefer calling `devices.get_device_for("controlnet")` instead of directly calling `get_optimal_device_name`.
I also think this is better, so I'll close it.

Issue #2506, opened 2024-01-19
[Bug]: RuntimeError: don't know how to restore data location of  torch.storage.UntypedStorage (tagged with privateuseone:0)


DrakeRowan:
### Is there an existing issue for this?

- [X] I have searched the existing issues and checked the recent builds/commits of both this extension and the webui

### What happened?

ControlNet is ignoring IP-Adapter (as well as T2I-Adapter) due to a runtime error, making outputs not take into account ControlNet at all even when it is enabled. No skeleton is shown either similar to the screens found [here](https://github.com/AUTOMATIC1111/stable-diffusion-webui/discussions/14571) in this likely related issue. Strangely however, the original ControlNet functions (i.e Depth/OpenPose/etc...) and their models from lllyasviel are working fine.

This is coming after a fresh install of StabilityMatrix as well as SD.Next Web UI.

IP-Adapter was working fine for me until a couple of days ago after I updated both ControlNet and IP-Adapter models. I presented the issue to SD.Next WebUI discord, but a user there pointed out that it was a ControlNet issue, thus now I am here. 

I do also want to add that this problem does not exist if --use-DirectML arguement is disabled and I allow SD to use CPU to generate images, but doing so increases it/s tenfold (of course). Again, ControlNet and IP-Adapter was working fine on my end just a couple of days ago for months, until the updates...

### Steps to reproduce the problem

1. Download and run Stability Matrix
2. Install SD.Next Web UI (by Vladmandic), Master ver via Stability Matrix
3. Launch SD.Next Web UI via Stability Matrix launcher /w --use-DirectML argument enabled
4. Latest ControlNet is already installed. Thus, once WebUI loads, click Image tab.
5. Upload an image, add prompts, change sampler, size, or w/e perimeters, etc...
6. Enable ControlNet (which, for me, is ControlNet v1.1.431)
7. Change ControlType to IP-Adapter, change preprocessor to IP-Adapter_clip_sd15
8. Download IP-Adapter_sd15.bin or .safetensors model from h94's huggingface, and place it in Stability Matrix's ControlNet model directory folder. (The ControlNet Models folder should be in the Models folder where StabilityMatrix.exe is after it was installed).
9. Go back to WebUI, and refresh ControlNet. The recently downloaded model should appear in dropdown. 
10. Attempt to Generate an Image.

### What should have happened?

[...]

PR #2503, opened 2024-01-19
 Apply a small patch(devices.get_device_for)

kingljl:
I have already closed the previous one(Address：https://github.com/Mikubill/sd-webui-controlnet/pull/2500/files ) and brought up a new one。

Follow your suggestion and change to devices.get_device_for("controlnet") method。
If you have already made changes, I will close this.   thank！！！

PR #2508, opened 2024-01-20
Revert " Apply a small patch(devices.get_device_for)"

huchenlei:
Reverts Mikubill/sd-webui-controlnet#2503.

Reason: Break AMD card users (#2506).

I am not very sure how to approach this as I do not own an AMD card. Maybe the change in lama/zoe are fine, but this needs further verification.

### DIFF
--- a/annotator/clipvision/__init__.py
+++ b/annotator/clipvision/__init__.py
@@ -79,10 +79,10 @@ downloads = {
 
 
 clip_vision_h_uc = os.path.join(os.path.dirname(os.path.abspath(__file__)), 'clip_vision_h_uc.data')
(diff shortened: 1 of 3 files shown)


kingljl:
> Reverts #2503.
> 
> Reason: Break AMD card users (#2506).
> 
> I am not very sure how to approach this as I do not own an AMD card. Maybe the change in lama/zoe are fine, but this needs further verification.

I have multiple GPU card requirements now, and I need to use devices. get_device_for ("controllnet") to solve the issue of 'cuda' if torch. cuda. is-available() else 'CPU'. The last modification was rolled back. When will we repair this piece?

Background of demand:

1. 4GPU card machine

2. Start 4 Diffusion processes using different cards

Issue #2494, opened 2024-01-18
[Bug]: Reordering the UI and placing the controlnet box before the dimensions box breaks the UI


emorris00:
### Is there an existing issue for this?

- [X] I have searched the existing issues and checked the recent builds/commits of both this extension and the webui

### What happened?

If you change the 'UI item order for txt2img/img2img tabs' setting and place the dimensions after scripts then the UI breaks and you have to manually change the config.json to fix it.

### Steps to reproduce the problem

1. Go to Settings > User Interface > User Interface
2. Change 'UI item order for txt2img/img2img tabs' settings and place dimensions after scripts
3. Apply Settings
4. Reload UI

### What should have happened?

The UI should not completely break.

### Commit where the problem happens

webui: cf2772fab0af5573da775e7437e6acdca424f26e
controlnet: ba05e1ea20df0e29a3dce9fcbe3ecc30777a5271

### What browsers do you use to access the UI ?

Mozilla Firefox, Microsoft Edge

### Command Line Arguments

```Shell
--medvram-sdxl --xformers --opt-split-attention
```

### List of enabled extensions

LDSR
Lora
ScuNET
SwinIR
canvas-zoom-and-pan
extra-options-section
hypertile
mobile
prompt-bracket-checker
Stable-Diffusion-Webui-Civitai-Helper
openpose-editor
sd-webui-controlnet
sd-webui-lobe-theme
stable-diffusion-webui-images-browser

### Console logs

```Shell
C:\coding\stable-diffusion>webui-user.bat
venv "C:\coding\stable-diffusion\venv\Scripts\Python.exe"
Python 3.10.11 (tags/v3.10.11:7d4cc5a, Apr  5 2023, 00:38:17) [MSC v.1929 64 bit (AMD64)]
Version: v1.7.0
Commit hash: cf2772fab0af5573da775e7437e6acdca424f26e
Launching Web UI with arguments: --medvram-sdxl --xformers --opt-split-attention
Style database not found: C:\coding\stable-diffusion\styles.csv
Civitai Helper: Get Custom Model Folder
ControlNet preprocessor location: C:\coding\stable-diffusion\extensions\sd-webui-controlnet\annotator\downloads
2024-01-18 15:17:14,325 - ControlNet - INFO - ControlNet v1.1.431
2024-01-18 15:17:14,394 - ControlNet - INFO - ControlNet v1.1.431
[...]

PR #2509, opened 2024-01-20
Update model dropdown choices when sd checkpoint change

### LINKED ISSUE #2270: [DevTask] Update ControlNet model list when SD model changes
See #2267.

Currently the ControlNet model list will be outdated when SD model changes. We need to trigger an update when SD model changes in A1111.

huchenlei:
Closes https://github.com/Mikubill/sd-webui-controlnet/issues/2494.
Closes #2270.

This PR adds the callback when sd checkpoint config. When it changes, model choices will be updated to match the current sd version.

This PR also does a large refactoring to delay register of callbacks after all A1111 component that we use/care have been initialized.

Local testing has verified that following features are still are working:
- [x] loopback checkbox visibility
- [x] batch dir sync
- [x] hr_option visibility
- [x] crop input image visibility
- [x] setting sd model checkpoint
- [x] send dimensions

### DIFF
--- a/scripts/controlnet.py
+++ b/scripts/controlnet.py
@@ -238,14 +238,12 @@ class Script(scripts.Script, metaclass=(
 
     def uigroup(self, tabname: str, is_img2img: bool, elem_id_tabname: str, photopea: Optional[Photopea]) -> Tuple[ControlNetUiGroup, gr.State]:
         group = ControlNetUiGroup(
+            is_img2img,
             Script.get_default_ui_unit(),
             self.preprocessor,
             photopea,
         )
-        group.render(tabname, elem_id_tabname, is_img2img)
-        group.register_callbacks(is_img2img)
-        unit_state = group.render_and_register_unit(tabname, is_img2img)
-        return group, unit_state
+        return group, group.render(tabname, elem_id_tabname)
 
     def ui_batch_options(self, is_img2img: bool, elem_id_tabname: str):
         batch_option = gr.Radio(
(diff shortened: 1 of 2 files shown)


emorris00:
Seems there are 2 problems with this. The first is that the callbacks don't get re-registered when the UI is reloaded using the Reload UI button. Second is that the sd checkpoint change updating the models dropdown only works when the 'Checkpoint dropdown: use filenames without paths' setting is set.

huchenlei:
@emorris00 
Thanks for spotting these issues! I have fixed the first UI reload issue. For the second issue, I cannot reproduce it. I do not have `Checkpoint dropdown: use filenames without paths` config enabled, but still checkpoint change indeed cause model list to update (You need to select any control type other than "All").

emorris00:
@huchenlei I have the open pose control type selected. When I don't have that setting set the change event fires immediately when the checkpoint dropdown is changed but the checkpoint hasn't actually changed yet so the model dropdown gets filtered base on the previous checkpoint. When the setting is selected it still does that, but then the change event get fired again after the checkpoint actually changes (the checkpoint name in the dropdown is changed to include the extension) so it then filters correctly based on the current checkpoint. I disabled all extensions except this and running on windows 11 with python 3.10.11.

huchenlei:
Ok, I successfully reproduced the issue. It seems like it only happens if you directly use the dropdown to select checkpoint. If you use the checkpoint tab with thumbnail preview to pick checkpoint, there won't be any issue.

PR #2533, opened 2024-01-21
:art: Multi-Inputs more user friendly UI

### LINKED ISSUE #2532: Consider using a more intuitive multi-image approach?
image

In general, within ControlNet, users wouldn't typically need to use more than 10 images, right? Can we consider adopting the upload method shown in the image below? This way, there's no need to go through the tedious process of organizing images and then copying and pasting paths.

PixPin_2024-01-22_03-11-34

huchenlei:
Closes #2532.

### DIFF
--- a/scripts/controlnet.py
+++ b/scripts/controlnet.py
@@ -388,8 +388,7 @@ class Script(scripts.Script, metaclass=(
                 return None
 
         attribute_value = get_element(getattr(p, attribute, None), strict)
-        default_value = get_element(default)
-        return attribute_value if attribute_value is not None else default_value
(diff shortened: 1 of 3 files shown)


scripts/controlnet.py
```diff
@@ -388,8 +388,7 @@ def get_element(obj, strict=False):
                 return None
 
         attribute_value = get_element(getattr(p, attribute, None), strict)
-        default_value = get_element(default)
```
huchenlei, line 391: The default value here now can be list. Calling `get_element` will get the item on `idx`, which is not what we want.

Generally I don't think here `get_element` call is justified on `default`.

Issue #2283, opened 2023-11-23
save ip adapter preprocessor result


godsgo:
Hello, thanks for your big job done.
I have a question can i save preprocessor result of image? Like we can do with pose or canny, and take it back to controlnet.
For example for creating ipdapter wildcards with saving time for generation without preprocessor
preprocessoring of image do 30-40% time of generation for first time, then it takes 1-2%

huchenlei:
I would say it is a valid request. Currently we don't offer download of binary preprocessor results. It is not just IP Adapter, but also for revision control type.

However, the code we currently have assumes the input to be image in many places, so it would be a very large effort to plumb the change through. Does this become a bottleneck for your workflow?

godsgo:
thanks for your answer, I made a style generator telegram bot with image prompts, this is method is fast for develop new style. Fast style creation - slow generation )

i saw in code cpu usage, maybe i mistake, but if we have chose gpu this can help doing job faster...

class IPAdapterModel(torch.nn.Module):
    def __init__(self, state_dict, clip_embeddings_dim, cross_attention_dim, is_plus, sdxl_plus, is_full):
        super().__init__()
        self.device = "cpu"

I dont readed all code just part, i am not specialist, just thoughts out loud. If its hard to emplement, better look on other enchantments. For me all work perfect, i am just ask. thanks in advance...

huchenlei:
Now we allow user run CLIP preprocessor on CUDA, hopefully the need to save CLIP result becomes unnecessary now. Please reopen if you would still like CLIP result to be downloaded/uploaded.

godsgo:
thank you very much it became much faster

PR #2590, opened 2024-01-27
:sparkles: Save/Use CLIP output in API

huchenlei:
Closes https://github.com/Mikubill/sd-webui-controlnet/issues/2283.

This PR allows user save CLIP output by calling `controlnet/detect` API, and use them in API payload.

Detect Payload example in tests/web_api/ipadapter_clip_api.py:
```python
def test_ipadapter_clip_api():
    """Use previously saved CLIP output in ipadapter run."""
    resp = detect_template(
        dict(
            controlnet_input_images=[realistic_girl_face_img],
            controlnet_module="ip-adapter_clip_h",
        )
    )
    ipadapter_input = resp["tensor"]
    APITestTemplate(
        "test_ipadapter_clip_api",
        "txt2img",
        payload_overrides={},
        unit_overrides={
            "ipadapter_input": ipadapter_input,
            "model": get_model("ip-adapter_sd15"),
        },
    ).exec()

```

### DIFF
--- a/internal_controlnet/external_code.py
+++ b/internal_controlnet/external_code.py
@@ -413,9 +413,6 @@ def to_processing_unit(unit: Union[Dict[str, Any], ControlNetUnit]) -> ControlNe
 
             unit["ipadapter_input"] = [decode_base64(b) for b in unit["ipadapter_input"]]
 
-            if unit["image"] is not None:
-                logger.warn(f"ipadapter_input detected in unit. image field will be ignored.")
-
         if 'guess_mode' in unit:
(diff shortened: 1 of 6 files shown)


vinch00:
+1 for this feature.
It would save so much time in loading IP Adapters.

With the current Turbo and Lightning models + LCM, generation is really fast, but loading IP Adapter can be slow due to CLIP outputs that need to be generated each time. If the CLIP outputs are saved from before and loaded with the API it will make generation enormously faster.

Issue #2593, opened 2024-01-28
[Bug]: img-to-img Unit 1 wont let me select Control Type, mouse icon turn into a prohibition icon


fifskank:
### Is there an existing issue for this?

- [X] I have searched the existing issues and checked the recent builds/commits of both this extension and the webui

### What happened?

in imgtoimg Unit 0 is enabled and i can select Control Type, but when i got to Unit 1, the mouse turns into prohibition icon in Control Type and wont let me select any

### Steps to reproduce the problem

1. Go to .... imgtoimg Enable Unit 0, select Control Type, then go to Unit 1 and Enable it and try to select a Control Type
2. Press .... the Control Type section but mouse turns in prohibition icon
3. ...

### What should have happened?

enable Unit 0 and select Control Type, Tile/Blur, then go to Unit 1 and select Control Type Canny,

### Commit where the problem happens

multi controlnet units only works Unit 0, Unit 1 cant select Control Type

### What browsers do you use to access the UI ?

Mozilla Firefox

### Command Line Arguments

```Shell
im using directml version for AMD
```

### List of enabled extensions

only ControlNet

### Console logs

```Shell
2024-01-28 00:41:54,190 - ControlNet - INFO - unit_separate = False, style_align = False02/153 [11:15<05:23,  6.34s/it]
2024-01-28 00:41:54,191 - ControlNet - WARNING - [tile_resample.threshold_a] Invalid value(-1), using default value 1.0.
2024-01-28 00:41:54,192 - ControlNet - INFO - Loading model from cache: control_v11f1e_sd15_tile [a371b31b]
2024-01-28 00:41:54,232 - ControlNet - INFO - Using preprocessor: tile_resample
2024-01-28 00:41:54,232 - ControlNet - INFO - preprocessor resolution = -1
2024-01-28 00:41:54,262 - ControlNet - INFO - ControlNet Hooked - Time = 0.07406759262084961
100%|██████████████████████████████████████████████████████████████████████████████████| 17/17 [01:48<00:00,  6.37s/it]
2024-01-28 00:43:48,410 - ControlNet - INFO - unit_separate = False, style_align = False19/153 [13:09<03:35,  6.34s/it]
2024-01-28 00:43:48,410 - ControlNet - WARNING - [tile_resample.threshold_a] Invalid value(-1), using default value 1.0.
[...]

PR #2595, opened 2024-01-28
:bug: Disable checkpoint callback for SDNext

huchenlei:
Closes https://github.com/vladmandic/automatic/issues/2760.

SDNext users have reported that ControlNet UI callbacks are not registered. After some investigation, we found that `setting_sd_model_checkpoint` is not available in SDNext. This PR disabled callback currently bind on `setting_sd_model_checkpoint` until we find a good way to register callbacks after all gradio components are ready in A1111. Currently `modules/script_callbacks.py` does not provide such hook. There is a "on_before_ui" but no "on_after_ui" now.

### DIFF
--- a/scripts/controlnet_ui/controlnet_ui_group.py
+++ b/scripts/controlnet_ui/controlnet_ui_group.py
@@ -74,14 +74,18 @@ class A1111Context:
 
     @property
     def ui_initialized(self) -> bool:
-        # Optional components are only available after A1111 v1.7.0.
         optional_components = {
+            # Optional components are only available after A1111 v1.7.0.
             "img2img_img2img_tab": "img2img_img2img_tab",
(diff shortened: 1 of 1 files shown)


vladmandic:
i've added on_after_ui callback if you want to use it (but should check if it exists since a1111 does not have it nor do older versions of sdnext)

huchenlei:
Thanks for implementing that callback! I think we just need to wait a while until the API is rolled to enough of population before adpoting it.

According to https://github.com/Mikubill/sd-webui-controlnet/issues/2593#issuecomment-1914825424, https://github.com/Mikubill/sd-webui-controlnet/pull/2597 has the issue fixed in SDNext for now.

vladmandic:
yes, i've tested it - latest controlnet now works with sdnext.
regarding on_after_ui callback, you'll probably have to submit a pr to a1111 to get it adopted?

PR #2646, opened 2024-02-16
Correct image saving in API examples

missionfloyd:
Using PIL to save images re-encodes them, losing metadata and potentially quality (if using lossy compression like JPEG). This way, images are saved as-is, and are identical to those saved by webui.

The wiki page should probably be updated as well.

Example:
_(Open in PNG Info)_
| Before | After |
|--------|--------|
| img2img-0 | img2img-0 |


huchenlei (approved):
LGTM

PR #2649, opened 2024-02-18
Compatibility fix for WebUI < v1.6.0

h3rmit-git:
Added fallback detection of component "txt2img_hr-checkbox" for WebUI versions < 1.6.0.

### DIFF
--- a/scripts/controlnet_ui/controlnet_ui_group.py
+++ b/scripts/controlnet_ui/controlnet_ui_group.py
@@ -111,6 +111,8 @@ class A1111Context:
             "img2img_inpaint_upload_tab": "img2img_inpaint_upload_tab",
             "img2img_inpaint_full_res": "img2img_inpaint_area",
             "txt2img_hr-checkbox": "txt2img_enable_hr",
+            # backward compatibility for webui < 1.6.0
+            "txt2img_enable_hr": "txt2img_enable_hr",
             # setting_sd_model_checkpoint is expected to be initialized last.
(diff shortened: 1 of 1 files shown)


huchenlei:
I would say better add `txt2img_enable_hr-checkbox` to `optional_components` in
```python
    @property
    def ui_initialized(self) -> bool:
        optional_components = {
            # Optional components are only available after A1111 v1.7.0.
            "img2img_img2img_tab": "img2img_img2img_tab",
            "img2img_img2img_sketch_tab": "img2img_img2img_sketch_tab",
            "img2img_batch_tab": "img2img_batch_tab",
            "img2img_inpaint_tab": "img2img_inpaint_tab",
            "img2img_inpaint_sketch_tab": "img2img_inpaint_sketch_tab",
            "img2img_inpaint_upload_tab": "img2img_inpaint_upload_tab",
            # SDNext does not have this field. Temporarily disable the callback on
            # the checkpoint change until we find a way to register an event when
            # all A1111 UI components are ready.
            "setting_sd_model_checkpoint": "setting_sd_model_checkpoint",
        }
        return all(
            c
            for name, c in vars(self).items()
            if name not in optional_components.values()
        )
```

h3rmit-git:
Adding it in optional components would prevent the registration of a callback event when hires.fix is changed in versions < 1.6.0. 

By adding a fallback instead, register_shift_hr_options() properly registers the callback event in those versions.

Tested in WebUI v1.3.0.

Issue #2113, opened 2023-09-13
[Bug]: TypeError: expected size to be one of int or Tuple[int] or Tuple[int, int] or Tuple[int, int, int]


Enferlain:
### Is there an existing issue for this?

- [X] I have searched the existing issues and checked the recent builds/commits of both this extension and the webui

### What happened?

Trying to use depth zoe with an image and pressing generate after enabling cn throws an error that I've never seen.

all the other preprocessors seem to work fine, seems like it's just depth zoe for some reason

torch: 2.2.0.dev20230911+rocm5.6 

### Steps to reproduce the problem

1. Be on amd
2. Follow the steps on the extension page 
3. generate or run preprocessor with depth_zoe selected

### What should have happened?

No errors

### Commit where the problem happens

webui: https://github.com/AUTOMATIC1111/stable-diffusion-webui/commit/5ef669de080814067961f28357256e8fe27544f4
controlnet: https://github.com/Mikubill/sd-webui-controlnet/commit/e67e017731aad05796b9615dc6eadce911298ea1

### Command Line Arguments

```Shell
None
```

### List of enabled extensions

![image](https://github.com/Mikubill/sd-webui-controlnet/assets/15861396/2495410e-9e03-4792-96e3-f064d8a78ee6)

### Console logs

```Shell
*** Error running process: /home/imi/stable-diffusion-webui/extensions/sd-webui-controlnet/scripts/controlnet.py
    Traceback (most recent call last):
      File "/home/imi/stable-diffusion-webui/modules/scripts.py", line 619, in process
        script.process(p, *script_args)
      File "/home/imi/stable-diffusion-webui/extensions/sd-webui-controlnet/scripts/controlnet.py", line 977, in process
        self.controlnet_hack(p)
      File "/home/imi/stable-diffusion-webui/extensions/sd-webui-controlnet/scripts/controlnet.py", line 966, in controlnet_hack
        self.controlnet_main_entry(p)
      File "/home/imi/stable-diffusion-webui/extensions/sd-webui-controlnet/scripts/controlnet.py", line 808, in controlnet_main_entry
        detected_map, is_image = preprocessor(
                                 ^^^^^^^^^^^^^
      File "/home/imi/stable-diffusion-webui/extensions/sd-webui-controlnet/scripts/utils.py", line 75, in decorated_func
        return cached_func(*args, **kwargs)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #2650, opened 2024-02-18
type error in depth zoe, converting numpy type into int

Nashihikari:
This issue still occurs in pytorch 2.1.2. As a library with such a broad user base, if the cost is not too high, I think we should consider compatibility issues to ensure that our users do not feel lost when encountering these errors.
#2113


huchenlei (approved):

huchenlei:
LGTM

hben35096:
This problem also exists in forge

PR #2352, opened 2023-12-24
[WIP] AnimateDiff V2.0.0 ControlNet Part

continue-revolution:
- [ ] https://github.com/continue-revolution/sd-webui-animatediff/pull/381

PR #2661, opened 2024-02-26
AnimateDiff 2.0.0-a ControlNet part

continue-revolution:
Given that huchenlei has implemented multi-image upload, I don't have to make too much change to the code now.

New features will only include SparseCtrl. Magic Animate and all other new works that include reference models will be forge only.

We potentially can and need to re-write this extension in a way similar to forge so that we can make maintenance easier.

#2352 will be abandoned. I expect to merge this PR in this week.

### DIFF
--- a/scripts/controlnet.py
+++ b/scripts/controlnet.py
@@ -3,7 +3,7 @@ import tracemalloc
 import os
 import logging
 from collections import OrderedDict
-from copy import copy
+from copy import copy, deepcopy
 from typing import Dict, Optional, Tuple, List, Union
 import modules.scripts as scripts
 from modules import shared, devices, script_callbacks, processing, masking, images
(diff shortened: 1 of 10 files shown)


scripts/controlnet.py
```diff
@@ -635,6 +669,14 @@ def decode_image(img) -> np.ndarray:
                 # Add mask logic if later there is a processor that accepts mask
                 # on multiple inputs.
                 input_image = [HWC3(decode_image(img['image'])) for img in image]
+                if unit_is_ad_batch and len(image) > 0 and 'mask' in image[0] and image[0]['mask'] is not None:
+                    for idx in range(len(input_image)):
+                        while len(image[idx]['mask'].shape) < 3:
+                            image[idx]['mask'] = image[idx]['mask'][..., np.newaxis]
+                        if 'inpaint' in unit.module:
```
huchenlei, line 676: Is `ad_batch` used only for ControlNet inpaint?

scripts/controlnet.py
```diff
@@ -894,29 +936,36 @@ def controlnet_main_entry(self, p):
                 model_net, control_model_type = Script.load_control_model(p, unet, unit.model)
                 model_net.reset()
 
+                if model_net is not None and getattr(devices, "fp8", False) and not isinstance(model_net, PlugableIPAdapter):
```
huchenlei, line 939: This logic is applied to all existing ControlNet usages. Do we want to always enable fp8 when it is available?

huchenlei (commented):
Is there link to guide on how to use the new feature with animate diff?

continue-revolution:
I write documentation for all features at the end of the update. I am implementing SparseCtrl and fixing IP-Adapter, so maybe need some more time

scripts/controlnet.py
continue-revolution: iirc this piece of code only applies to ad_batch. If not I will fix

continue-revolution: Yes it is. Your multi-input implementation also access this. I don’t other features of CN access this part. b.t.w., your multi-input implementation is very helpful for me to remove CN hook, but I still think we need to find a way to re-implement CN in some way like forge, to make our life easier

scripts/controlnet.py
continue-revolution: I’m not sure. But from my experiments I think fp8 is ok when CN+AD+A1111 are all in fp8

PR #2665, opened 2024-02-29
Update processor.py

George0726:
Fix a tiny bug for facial area calculation during selecting the largest facial area.


huchenlei (approved):
LGTM

PR #2666, opened 2024-02-29
Add ruff linter dep

huchenlei:
Add ruff linter dependency. We will gradually fix existing lint issues and get directories down from exclude list.

### DIFF
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -1,0 +1,24 @@
+name: Linter
+
+on:
+  - push
+  - pull_request
+
+jobs:
+  lint-python:
+    name: ruff
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout Code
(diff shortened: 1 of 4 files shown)


continue-revolution:
This linter will bring a tremendous amount of trouble to developers and I strongly urge to remove it.

continue-revolution:
alright I found a way to bypass ruff in some intentional lines. I found it is useful in some cases

huchenlei:
> alright I found a way to bypass ruff in some intentional lines. I found it is useful in some cases

I copied the config from A1111 main repo. I think AUTO already disabled most annoying checks. I find the remaining checks useful to improve code quality (For example, it will remind you unused import on PR). If you really need to bypass linter, you can either add `# noqa: {ERROR_CODE}` at the end of line to supress linter or simply ignore the file in `pyproject.toml`.

PR #2667, opened 2024-02-29
Fix linter issues under `/script`

huchenlei:
Fix many linter errors in under `/script`.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -8,7 +8,8 @@ exclude = [
   "web_tests",
   "example",
   "extract_controlnet_diff.py",
-  "scripts",
+  "scripts/global_state.py",
+  "scripts/movie2movie.py",
 ]
 
 ignore = [
--- a/scripts/adapter.py
+++ b/scripts/adapter.py
@@ -2,9 +2,8 @@ import torch
 import torch.nn as nn
 from collections import OrderedDict
 
-from omegaconf import OmegaConf
 from copy import deepcopy
-from modules import devices, lowvram, shared, scripts
+from modules import devices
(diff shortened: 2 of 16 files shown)


Myoko:
When i using inpaint with adetailer, I got an error message "diffusionwrapper.forward() missing 2 required positional arguments: 'x' and 't'".

huchenlei:
> When i using inpaint model in adetailer, I got an error message "diffusionwrapper.forward() missing 2 required positional arguments: 'x' and 't'".

Please file a separate issue. I do not think this PR causes the issue as no function signature has changed.

Myoko:
> > When i using inpaint model in adetailer, I got an error message "diffusionwrapper.forward() missing 2 required positional arguments: 'x' and 't'".
> 
> Please file a separate issue. I do not think this PR causes the issue as no function signature has changed.

I'm not sure if this PR is the cause of the problem, i trying to roll back PR#2666 and everything is normal.

huchenlei:
I think you can ignore this error for now. I think the error was previously discarded silently here:
```python
                try:
                    # Trigger the register_forward_pre_hook
                    outer.sd_ldm.model()
                except Exception as e:
                    logger.error(e)
```

Now after the change we just report it.

I will change the log level for that log to `debug` so you won't see it normally.

Myoko:
> You can ignore this error for now, the functionality is unchanged. I think the error was previously discarded silently here:
> 
> ```python
>                 try:
>                     # Trigger the register_forward_pre_hook
>                     outer.sd_ldm.model()
>                 except Exception as e:
>                     logger.error(e)
> ```
> 
> Now after the change we just report it.
> 
> I will change the log level for that log to `debug` so you won't see it normally.

I thought this error loading inpaint model failed! OK, thank you.

PR #2668, opened 2024-02-29
make it compitable with other torch devices which support fp16 dtype

python279:
1. specify dtype for specific devices
2. convert the depth to float after model forward


huchenlei:
Can you elaborate on what this PR is trying to solve? i.e. before this PR, was there an issue with midas annotator?

python279:
> Can you elaborate on what this PR is trying to solve? i.e. before this PR, was there an issue with midas annotator?

We encountered the torch aten did not support fp32 on our NPU, we set the default dtype to torch.float16 in weibui devices.py, suppose the extentions should follow the configuration in webui.

Issue #2800, opened 2024-04-23
Keep active preprocessor modules loaded between runs


huchenlei:
> > SDXL ipadapter uses clip-g as preprocessor which is slower than clip-h. You can try switch to `ip-adapter_sdxl_vit-h` which uses clip-h as preprocessor.
> 
[...]

PR #2754, opened 2024-04-16
Port SDForge's preprocessor structure

huchenlei:
Closes https://github.com/Mikubill/sd-webui-controlnet/issues/2748
Closes #2419

- [x] Add back sparse ctrl
- [x] Add back ip-adapter-auto
- [x] Add back preprocessor cache
- [x] Add optional Marigold depth preprocessor

Marigold depth preprocessor added https://github.com/huchenlei/sd-webui-controlnet-marigold

User visible change:
Control type radio order will be changed to alphabetical order:
image

PR #2814, opened 2024-04-27
Keep active preprocessor models

huchenlei:
Closes https://github.com/Mikubill/sd-webui-controlnet/issues/2800.

Preprocessor model unload behaviour changed in https://github.com/Mikubill/sd-webui-controlnet/pull/2754 to align with SD forge's impl. Forge does eager unload, which is not ideal for CLIP, which takes very long to load/unload. This PR restores the original behaviour which keeps active preprocessors between runs.

### DIFF
--- a/internal_controlnet/external_code.py
+++ b/internal_controlnet/external_code.py
@@ -283,6 +283,13 @@ class ControlNetUnit:
                     f"[{self.module}.{unit_param}] Invalid value({value}), using default value {cfg.value}."
                 )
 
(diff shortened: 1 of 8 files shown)


Dasor92:
I don't know if it's caused by this, but changing preprocessors now doesn't unload the previous one and now if you use PULID, then another different input always with PULID it adds 3gb of VRAM used, and so on, i reached 40gb of Vram used in just one session just by changing the photo input

Issue #2836, opened 2024-05-03
[Bug]: Error on install.py (ModuleNotFoundError: No module named 'launch')


shsotoudegan:
### Is there an existing issue for this?

- [x] I have searched the existing issues and checked the recent builds/commits of both this extension and the webui

### What happened?

I am deploying A1111 on replicate.com with cog
i was clone ControlNet repository in extentions directory of A1111.
other extentions are installed except ControlNet

### Steps to reproduce the problem

1. put `git clone https://github.com/Mikubill/sd-webui-controlnet.git /stable-diffusion-webui/extensions/sd-webui-controlnet` on cog.yam; build -> run
2. run `cog push` for push on replicate.com
3. get this error:
`
#24 14.40 Installing requirements
#24 49.46 *** Error running install.py for extension /stable-diffusion-webui/extensions/sd-webui-controlnet.
#24 49.46 *** Command: "/root/.pyenv/versions/3.10.12/bin/python" "/stable-diffusion-webui/extensions/sd-webui-controlnet/install.py"
#24 49.46 *** Error code: 1
#24 49.46 *** stderr: Traceback (most recent call last):
#24 49.46 ***   File "/stable-diffusion-webui/extensions/sd-webui-controlnet/install.py", line 1, in <module>
#24 49.46 ***     import launch
#24 49.46 *** ModuleNotFoundError: No module named 'launch'
`

### What should have happened?

should this install on A1111

### Commit where the problem happens

webui: last
controlnet: last

### Command Line Arguments

```Shell
cog push
```

### List of enabled extensions

adetailer

### Console logs

```Shell
#24 14.40 Installing requirements
#24 49.46 *** Error running install.py for extension /stable-diffusion-webui/extensions/sd-webui-controlnet.
#24 49.46 *** Command: "/root/.pyenv/versions/3.10.12/bin/python" "/stable-diffusion-webui/extensions/sd-webui-controlnet/install.py"
#24 49.46 *** Error code: 1
#24 49.46 *** stderr: Traceback (most recent call last):
#24 49.46 ***   File "/stable-diffusion-webui/extensions/sd-webui-controlnet/install.py", line 1, in <module>
#24 49.46 ***     import launch
#24 49.46 *** ModuleNotFoundError: No module named 'launch'
```

### Additional information

cog.yaml (forked from https://github.com/speshiou/cog-sd-webui)
```yaml
# Configuration for Cog ⚙️
[...]

PR #2839, opened 2024-05-03
fix: installation. launch ModuleNotFoundError

shsotoudegan:
This pull request addresses the issue described in #2836 . In some cases, the `launch` module in the webui fails to resolve because it's a relative import beyond the top-level package. This fix corrects that problem.

### DIFF
--- a/install.py
+++ b/install.py
@@ -1,4 +1,3 @@
-import launch
 from importlib import metadata
 import sys
 import os
@@ -6,6 +5,21 @@ import shutil
 import platform
 from pathlib import Path
 from typing import Tuple, Optional
+
+try:
+    import launch
+except ModuleNotFoundError:
+    print("ControlNet cannot resolve the 'launch' module from the webui. Attempting to add the webui path to sys.path.")
+
+    # Add webui to interpreter path
+    webui_path = Path(__file__).parent.parent.parent.absolute().resolve()
+    sys.path.append(str(webui_path))
+
+    try:
+        import launch
+        print("Successfully resolved the 'launch' module.")
+    except ModuleNotFoundError:
+        raise ModuleNotFoundError("Error: Still unable to resolve the 'launch' module after modifying sys.path. Please check the project structure.")
 
 
 repo_root = Path(__file__).parent


huchenlei (commented):
I think this environment issue should be fixed in A1111 repo, instead of adding complexity here.

shsotoudegan:
but `import beyond the top-level package` is not allowed in python. we can accept this changes or add A1111 in this project dependencies. or create another module or lib for handle `launch` functionality.

huchenlei:
> but `relative import beyond the top-level package` is not allowed in python. we can accept this changes or add A1111 to this project dependencies. or create another module or lib for handle `launch` functionality.

Please file an issue in A1111 repo. I think you need to adjust A1111's extension loading code to foundamentally fix this.

huchenlei:
> but `relative import beyond the top-level package` is not allowed in python. we can accept this changes or add A1111 to this project dependencies. or create another module or lib for handle `launch` functionality.

Relevant code is here: https://github.com/AUTOMATIC1111/stable-diffusion-webui/blob/1c0a0c4c26f78c32095ebc7f8af82f5c04fca8c0/modules/launch_utils.py#L227-L240

install.py
```diff
@@ -7,6 +6,21 @@
 from pathlib import Path
 from typing import Tuple, Optional
 
+try:
+    import launch
+except ModuleNotFoundError:
+    print("ControlNet cannot resolve the 'launch' module from the webui. Attempting to add the webui path to sys.path.")
+
+    # Add webui to interpreter path
+    webui_path = Path(__file__).parent.parent.parent.absolute().resolve()
```
w-e-w, line 15: this is also not a valid fix
extension directory may not be under webui root
it can be moved by [`--data-dir`](https://github.com/AUTOMATIC1111/stable-diffusion-webui/wiki/Command-Line-Arguments-and-Settings)

shsotoudegan: that's true,
I working with stable diffusion environment for only 4 days and for only a small project.
I don't know how it is work from scratch.
can I ask for help from both of you? @w-e-w @huchenlei

PR #2881, opened 2024-05-10
Fix upscale script checking

seancheung:
If Hires fix is enabled, controlnet inpu image will always be cropped since Hires fix's script name is `Hires Upscale(r)`.
This causes issues with some extensions like ADetailer. 
One case is when using IP-Adapter FaceID with ADetailer, the input face image will be cropped, and Insightface will fail to detect faces in the cropped image.

Screenshot 2024-05-11 052216

This pull request fixes this issue.

For future improvement, I suggest that a new argument (e.g. `is_cropping_disabled` with default value `False`) should be added to allow third-party extension to disable this cropping feature.

image


huchenlei (approved):
Lgtm

PR #2909, opened 2024-05-19
Fix ControlNet batch

### LINKED ISSUE #2886: [Bug]: Batch mode skip half of the images

### What happened?

 if 1 one was taken 2nd one skip then 3rd one take and 4th one skip and goes on

try batch with 10 images and only 5 will produce

generate all 1o images

webui: 
controlnet: 

Google Chrome

### Command Line Arguments

```Shell
no
```

### List of enabled extensions

no

### Console logs

```Shell
no
```

### Additional information

_No response_

huchenlei:
Closes #2886.

The root cause of this issue is that both img2img script instance and txt2img script instance has access to the same group of enabled_units objects, and each batch iteration, `batch_tab_process_each` on both img2img script and txt2img script are invoked to advance the iterator on the same unit.

Previously units are copied on each call to `Script.get_enabled_units`, which is costly, but maintains the correctness. This PR solves the issue by distinguish txt2img and img2img script, which resolves the issue more fundamentally, and also avoiding copy cost of ControlNetUnit.

### DIFF
--- a/scripts/controlnet.py
+++ b/scripts/controlnet.py
@@ -307,6 +307,10 @@ class Script(scripts.Script, metaclass=(
         self.post_processors = []
         self.noise_modifier = None
         self.ui_batch_option_state = [BatchOption.DEFAULT.value, False]
+        # `Script` instance is created twice, once for img2img and once for txt2img.
+        # However, A1111 does not pass is_img2img to script constructor, so this field
+        # is initialized in `ui` method.
+        self.is_img2img = False
         batch_hijack.instance.process_batch_callbacks.append(self.batch_tab_process)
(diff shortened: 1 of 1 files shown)


fastjhone0:
> Closes #2886.
> 
> The root cause of this issue is that both img2img script instance and txt2img script instance has access to the same group of enabled_units objects, and each batch iteration, `batch_tab_process_each` on both img2img script and txt2img script are invoked to advance the iterator on the same unit.
> 
> Previously units are copied on each call to `Script.get_enabled_units`, which is costly, but maintains the correctness. This PR solves the issue by distinguish txt2img and img2img script, which resolves the issue more fundamentally, and also avoiding copy cost of ControlNetUnit.

Thank you so much for this fix

PR #2952, opened 2024-06-09
Fix issue with > in pip cmd causing file creation

Uminosachi:
### Problem Summary

When starting the Web UI, the `install.py` script reads `requirements.txt`. If a package version uses ">", it is treated as a redirection in `pip install`, creating an unintended file.

### Steps to Reproduce

Running `webui.bat` or `webui.sh` creates a file named "=4.8.0" due to "opencv-python>=4.8.0" in `requirements.txt` being misinterpreted.

### Fix

I enclosed the package and version specification in double quotes in `install.py`.

### DIFF
--- a/install.py
+++ b/install.py
@@ -38,7 +38,7 @@ def install_requirements(req_file):
                     installed_version = get_installed_version(package_name)
                     if installed_version != package_version:
                         launch.run_pip(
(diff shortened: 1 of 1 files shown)


huchenlei (approved):
LGTM

PR #2953, opened 2024-06-09
Add security warning for install from environment variable override URL

huchenlei:
The `os.environ` override of dep path is meant to be used by package author in China to serve users who has difficulties accessing github. Original PR: https://github.com/Mikubill/sd-webui-controlnet/pull/2514

However, this poses a security risk to install any package from internet if someone set these environment variables somewhere else. This PR adds an explicit warning message to warn user when URLs from these environs are used. 

@Akegarasu Do you think there is a better way to handle this on your side? Current mechanism is somewhat unsafe.

### DIFF
--- a/install.py
+++ b/install.py
@@ -78,16 +78,21 @@ def install_onnxruntime():
     Install onnxruntime or onnxruntime-gpu based on the availability of CUDA.
     onnxruntime and onnxruntime-gpu can not be installed together.
     """
-    if not launch.is_installed("onnxruntime") and not launch.is_installed("onnxruntime-gpu"):
-        import torch.cuda as cuda # torch import head to improve loading time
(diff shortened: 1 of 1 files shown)


Akegarasu:
Actually, I don't think this is a security issue worth worrying about, perhaps due to the exposure of some node of comfyui today that some people are overly concerned.

Firstly, within sd-webui, there are many commands that directly read and execute environment variables, such as TORCH_COMMAND, or XFORMERS_PACKAGE, among others. This is a method that everyone uses.

Secondly, if a malicious program is already capable of modifying the environment variables, it is highly likely that it does not need to use this method to install a malicious package.

wfjsw:
With access to environment variable, I can get all my way into your code and remove the warning before it is even executed, so it is not much safer than the current state.

huchenlei:
Thanks for the explanation! I guess we can keep it as it is.

PR #3000, opened 2024-07-15
Add mobile_sam with controlnet_aux

sdbds:


sdbds:
image
image

It seems OK.

PR #3075, opened 2024-12-23
Add element IDs to advanced weight control components

marcsyp:
This PR adds element IDs to the Weight Type dropdown and Weights textbox in the Advanced Weight Control section of ControlNet units. The IDs follow the existing naming pattern used throughout controlnet_ui_group.py.

The primary purpose is to enable compatibility with the Config Presets extension, allowing users to save and reload complex ControlNet configurations including advanced weight settings.

Changes:
- Added elem_id parameters to AdvancedWeightControl.render()
- Added elem_ids to weight_type dropdown and weight_editor textbox
- Maintained backward compatibility for cases where render is called without parameters

The changes to the dropdown have been tested and confirmed working with both ControlNet and Config Presets.  Changes to the advanced weights textbox will need to rely on custom code in the Config Presets to stop propagation of Weight Type events when both the Weight Type and Advanced Weights are updated as part of the same preset (otherwise the UI event for Weight Type overwrites the Advanced Weights).

### DIFF
--- a/scripts/controlnet_ui/advanced_weight_control.py
+++ b/scripts/controlnet_ui/advanced_weight_control.py
@@ -72,7 +72,7 @@ class AdvancedWeightControl:
         self.weight_editor = None
         self.weight_composition = None
 
-    def render(self):
+    def render(self, tabname=None, elem_id_tabname=None):
         with gr.Group(visible=False) as self.group:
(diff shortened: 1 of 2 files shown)


ice-fly:
Reviewed commit appears to enhance compatibility.

